PR #170, opened 2023-12-11
MAM-3452-performance-init-photopicker-when-needed

bnolens:
PhotoPicker is initialized during app launch (27ms) but should only be initalized right before usage.
https://linear.app/theblvd/issue/MAM-3452/performance-init-photopicker-when-needed

Screenshot 2023-12-11 at 13 08 23


rileyhCode (approved):
Looks good to me.

PR #171, opened 2023-12-11
MAM-3456-performance-dont-update-unreadindicator-on-scroll

bnolens:
Instead of updating the unreadIndicator count and visibility (17ms) in `scrollViewDidScroll`, update it in `tableView:willDisplay:forRowAt` instead.

Calling `getCurrentCellIndexPath` (10ms) and `LatestPill.isEnabled.setter` (9ms) are expensive operations.

https://linear.app/theblvd/issue/MAM-3456/performance-dont-update-unreadindicator-on-scroll

Screenshot 2023-12-11 at 15 06 21

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
@@ -480,6 +480,22 @@ extension NewsFeedViewController {
         } else if let cell = cell as? ActivityCardCell {
             cell.display()
         }
+        
+        if self.viewModel.getUnreadEnabled(forFeed: self.viewModel.type) {
+            let currentRow = indexPath.row
+            let unreadState = self.viewModel.getUnreadState(forFeed: self.viewModel.type)
+            let count = min(currentRow, unreadState.count)
+            self.viewModel.setUnreadState(count: count, enabled: true, forFeed: self.viewModel.type)
+            
+            switch self.viewModel.type {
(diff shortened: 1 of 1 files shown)


rileyhCode (commented):
This is being called *way* less often now, but I still see it called several times while scrolling. Does it make sense to move it to where it only executes once scrolling has stopped?

bnolens:
> This is being called _way_ less often now, but I still see it called several times while scrolling. Does it make sense to move it to where it only executes once scrolling has stopped?

I didn't do that because on activity and mentions we want the indicator to count down while scrolling. I could create separate cases for those feeds. I'd prefer to know if it has a real impact on performance though before adding this special case.

rileyhCode:
> > > This is being called _way_ less often now, but I still see it called several times while scrolling. Does it make sense to move it to where it only executes once scrolling has stopped?
> > 
> > 
> > I didn't do that because on activity and mentions we want the indicator to count down while scrolling. I could create separate cases for those feeds. I'd prefer to know if it has a real impact on performance though before adding this special case.
> 
> Moving it to `tableView:willDisplay:forRowAt` removed the need of calling `self.getCurrentCellIndexPath(tableView: self.tableView)?.row`. I'm not seeing the other calls in the time profiler anymore, so I think we're good. I'll do more profiling during the week and will keep an eye out for this. I'd create another ticket if more optimization is needed on this though. Wdyt?

OK; sounds good. 👍

rileyhCode (approved):

PR #172, opened 2023-12-11
MAM-3460-performance-create-the-profile-pic-badge-only-once

bnolens:
Instead of adding and removing the badge from the view on each configure call, add the badge once, and use isHidden to show/hide it.

https://linear.app/theblvd/issue/MAM-3460/performance-create-the-profile-pic-badge-only-once


rileyhCode (changes requested):
Looks good. We should also remove the badgeContraints ivar.

bnolens:
> Looks good. We should also remove the badgeContraints ivar.

I did remove it, didn't I?

rileyhCode:
> > Looks good. We should also remove the badgeContraints ivar.
> 
> I did remove it, didn't I?

Yes; sorry, somehow I missed it.

rileyhCode (approved):
LGTM

PR #174, opened 2023-12-11
MAM-3342-should-not-show-gear-in-upper-right-of-profile-when-onboarding

rileyhCode:
Only show the gear in the navbar if the user is past onboarding.

### DIFF
--- a/Mammoth/Screens/ProfileScreen/ProfileViewController.swift
+++ b/Mammoth/Screens/ProfileScreen/ProfileViewController.swift
@@ -171,13 +171,17 @@ private extension ProfileViewController {
     func setupUI() {
         
         self.view.backgroundColor = .custom.background
+
+        let isPastOnboarding: Bool = (AccountsManager.shared.currentAccount as? MastodonAcctData)?.wentThroughOnboarding ?? true
+
+        if isPastOnboarding {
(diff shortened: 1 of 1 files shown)


Mammoth/Screens/ProfileScreen/ProfileViewController.swift
```diff
@@ -171,13 +171,17 @@ private extension ProfileViewController {
     func setupUI() {
         
         self.view.backgroundColor = .custom.background
+
+        let isPastOnboarding: Bool = (AccountsManager.shared.currentAccount as? MastodonAcctData)?.wentThroughOnboarding ?? true
```
bnolens, line 175: Wondering if passing a `hasSettingButton: Bool = true` attribute in the custom `init` method would make more sense. It keeps the VC a little dumber, which is a better pattern if you ask me.

bnolens (approved):
Proposing 1 improvement, but LGTM

Mammoth/Screens/ProfileScreen/ProfileViewController.swift
rileyhCode: I originally started down this path, but found too many ways in which the user can get to the profile pic from onboarding (from any smart list, posts, replies, etc.) that it became cumbersome to forward the hasSettingButton everywhere, and got quite messy.  :-/

PR #175, opened 2023-12-11
MAM-3398-move-settings-out-of-avatar-drop-down-and-into-under-customi…

rileyhCode:
…ze-for-you

- remove the Settings… item from the accounts menu
- add the Settings… menu item to … menu


bnolens (approved):
LGTM

PR #177, opened 2023-12-12
MAM-3469-performance-only-parse-user-description-html-when-loading

bnolens:
...-profile

We only need to parse the description html on the profile screen (to make all links work correctly). In other places (UserCardCell) we use a stripped version of the description html and display that instead (faster)

https://linear.app/theblvd/issue/MAM-3469/performance-only-parse-user-description-html-when-loading-profile


rileyhCode (approved):
LGTM

PR #178, opened 2023-12-12
MAM-3470-performance-time-should-not-be-computed-property

bnolens:
In PostCardModel and ActivityCardModel we do not want time to be a computed property. Calling NSDateFormatter methods just-in-time seems to be expensive. Instead we set time once during init and update the value when the label is displayed using a custom time interval.

https://linear.app/theblvd/issue/MAM-3470/performance-time-should-not-be-computed-property

### DIFF
--- a/Mammoth/Screens/DetailScreen/DetailViewController.swift
+++ b/Mammoth/Screens/DetailScreen/DetailViewController.swift
@@ -182,13 +182,13 @@ class DetailViewController: UIViewController {
 extension DetailViewController: UITableViewDataSource, UITableViewDelegate {
     func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
         if let cell = cell as? PostCardCell {
-            cell.display()
+            cell.willDisplay()
         }
     }
     
     func tableView(_ tableView: UITableView, didEndDisplaying cell: UITableViewCell, forRowAt indexPath: IndexPath) {
         if let cell = cell as? PostCardCell {
-            cell.endDisplay()
+            cell.didEndDisplay()
         }
     }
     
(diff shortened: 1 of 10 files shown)


Mammoth/Screens/DetailScreen/DetailViewController.swift
```diff
@@ -180,6 +180,18 @@ class DetailViewController: UIViewController {
...
 extension DetailViewController: UITableViewDataSource, UITableViewDelegate {
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        if let cell = cell as? PostCardCell {
+            cell.display()
+        }
+    }
+    
+    func tableView(_ tableView: UITableView, didEndDisplaying cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        if let cell = cell as? PostCardCell {
+            cell.endDisplay()
```
rileyhCode, line 191: Consider renaming these to .willDisplay() and .didEndDisplaying()

Mammoth/Views/Cells/ActivityCardCell/ActivityCardHeader.swift
```diff
@@ -72,15 +73,24 @@ class ActivityCardHeader: UIView {
     private var activity: ActivityCardModel?
     public var onPress: PostCardButtonCallback?
     
+    private let runLoop = RunLoop.main
```
rileyhCode, line 76: Consider not having a runLoop ivar and just using RunLoop.main directly.

Mammoth/Views/Cells/PostCardCell/PostCardHeader.swift
```diff
@@ -123,20 +124,32 @@ class PostCardHeader: UIView {
     private var followButton: FollowButton?
     
     private var status: Status?
+    private var postCard: PostCardModel?
     private var type: PostCardHeaderTypes = .regular
     public var onPress: PostCardButtonCallback?
     
+    private let runLoop = RunLoop.main
```
rileyhCode, line 131: Consider not having a runLoop ivar and just using RunLoop.main directly.

rileyhCode (approved):
Added a couple of comments for consideration.

PR #187, opened 2023-12-14
MAM-3472-remove-opt-out-behavior-for-mammoth-in-onboarding

rileyhCode:
- add third onboarding pane for Follow Mammoth
- filter mammoth@moth.social out of the list of suggested accounts
- no longer default to following mammoth@moth.social (was opt-out)
- skip suggesting following Mammoth if we're certain the user is already following it


bnolens (approved):
Looks good!

PR #193, opened 2023-12-15
MAM-3509-hard-returns-in-onboarding-headlines

rileyhCode:
- change from UILabel to UITextView for more control
- add line breaks for iPhone (and convert breaks to spaces for iPad)
- update the InstrumentSerif font


bnolens (approved):
LGTM

PR #197, opened 2023-12-18
MAM-3521-performance-cache-cell-heights-and-associate-value-with

bnolens:
https://linear.app/theblvd/issue/MAM-3521/performance-cache-cell-heights-and-associate-value-with-postcardmodel

### DIFF
--- a/Mammoth/Models/PostCardModel.swift
+++ b/Mammoth/Models/PostCardModel.swift
@@ -99,6 +99,7 @@ final class PostCardModel {
     
     var imagePrefetchToken: SDWebImagePrefetchToken?
     var decodedImages: [String: UIImage?] = [:]
+    var cellHeight: CGFloat?
     
     enum FilterType {
         case warn(String)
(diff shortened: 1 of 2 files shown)


Mammoth/Models/PostCardModel.swift
```diff
@@ -99,6 +99,7 @@ final class PostCardModel {
     
     var imagePrefetchToken: SDWebImagePrefetchToken?
     var decodedImages: [String: UIImage?] = [:]
+    var cellHeight: CGFloat?
```
rileyhCode, line 102: Does this get cleared out for the iPad rotation case?

rileyhCode (changes requested):
A question about keeping/clearing the cellHeight during iPad rotation.

Mammoth/Models/PostCardModel.swift
bnolens: Because this value is only used for the estimated cell height, it seems ok to leave it like this. In the case we clear it, it will rely on a fixed number (310 px) instead. This value is the same for all orientation and screen heights. So I don't see the value in adding this extra logic to invalidate the cached height. Am I missing something?

rileyhCode: OK; that sounds good.

rileyhCode (approved):

PR #199, opened 2023-12-18
MAM-3561-investigate-account-language-setting-and-how-to-incorporate-it

rileyhCode:
We now pick the post language in this priority:
1. previously picked language (from prefs), if available
2. default post language from their mastodon profile if available
3. their iOS current language (should always be available)
4. english if all else fails

### DIFF
--- a/Mammoth/Managers/AccountsManager/AcctData.swift
+++ b/Mammoth/Managers/AccountsManager/AcctData.swift
@@ -108,6 +108,7 @@ struct MastodonAcctData: AcctDataType {
     let account: Account
     let instanceData: InstanceData
     let defaultPostVisibility: Visibility
+    let defaultPostingLanguage: String?
     let emoticons: [Emoji]
     var forYou: ForYouAccount
     var wentThroughOnboarding: Bool // false for accounts coming from 1.x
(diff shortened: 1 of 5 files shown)


Mammoth/Managers/AccountsManager/AcctData.swift
```diff
@@ -120,18 +121,20 @@ struct MastodonAcctData: AcctDataType {
...
         case instanceData
         case defaultPostVisibility
+        case defaultPostLanguage
         case emoticons
         case forYou
         case wentThroughOnboarding
     }
     
-    init(account: Account, instanceData: InstanceData, client: Client, defaultPostVisibility: Visibility, emoticons: [Emoji], forYou: ForYouAccount, uniqueID: String? = nil, wentThroughOnboarding: Bool = false) {
+    init(account: Account, instanceData: InstanceData, client: Client, defaultPostVisibility: Visibility, defaultPostingLanguage: String?, emoticons: [Emoji], forYou: ForYouAccount, uniqueID: String? = nil, wentThroughOnboarding: Bool = false) {
```
bnolens, line 130: Have you considered adding a default value  `defaultPostingLanguage: String? = nil`? That way `defaultPostingLanguage` is optional when initalizing `MastodonAcctData`.

bnolens (approved):
Please review my comment but LGTM

Mammoth/Managers/AccountsManager/AcctData.swift
rileyhCode: I did, but I wanted callers to specifically think through what they want here. (this is called in a few places that have specific needs regarding whether to keep/not defaults)

PR #200, opened 2023-12-18
MAM-3543-update-some-icons-in-the-composer-toolbar

rileyhCode:
- update icons to FontAwesome
- handle wider language codes
- draw to spec


bnolens (approved):
LGTM

Issue #155, opened 2023-12-07
using the wrong endpoint for fetching profiles.


derspyy:
mammoth is using the following structure to fetch profile data:

```http
GET /api/v2/search?q=@username@domain&resolve=true
```

this returns an array with possible search matches.

instead, for full username requests, it should use the webfinger endpoint:

```http
GET /api/v1/accounts/lookup?acct=@username@domain
```

which returns the profile itself, which is easier and more performant.

derspyy:
https://github.com/TheBLVD/mammoth/blob/0d7aec5bb7987cf59972ef21f88c3e7a3a837738/Mammoth/Managers/FollowManager.swift#L186-L199

https://github.com/TheBLVD/mammoth/blob/0d7aec5bb7987cf59972ef21f88c3e7a3a837738/Mammoth/Managers/FollowManager.swift#L225-L241

https://github.com/TheBLVD/mammoth/blob/0d7aec5bb7987cf59972ef21f88c3e7a3a837738/Mammoth/Managers/FollowManager.swift#L353-L369

https://github.com/TheBLVD/mammoth/blob/0d7aec5bb7987cf59972ef21f88c3e7a3a837738/Mammoth/Managers/FollowManager.swift#L535-L548

assuming every time `Search.search()` is called the first account is picked, you'd do better with the lookup address which gives you the profile directly (which seems to be `Accounts.lookup()`?)

derspyy:
gotta stop calling stuff wrong

bvon:
@jtomchak & @bnolens, tagging you here to give this a look, as it relates to our discussions Friday re: performance

PR #204, opened 2023-12-19
MAM-3566-profile-fetching-could-use-a-more-performant-function

bnolens:
When doing an account look up, use the Account lookup endpoint instead of the generic search endpoint.

Linear: https://linear.app/theblvd/issue/MAM-3566/profile-fetching-could-use-a-more-performant-function
Github Issues: https://github.com/TheBLVD/mammoth/issues/155


rileyhCode (approved):
With resolve=true, some of these called used to ask the server to use webfinger if necessary. Just something to be aware of in case we see failures.

PR #207, opened 2023-12-19
MAM-3569-profile-avatar-image-reloads-after-already-being-loaded

bnolens:
https://linear.app/theblvd/issue/MAM-3569/profile-avatar-image-reloads-after-already-being-loaded


rileyhCode (approved):
LGTM

PR #210, opened 2023-12-20
MAM-3527-translation-not-working-build-v-202-build-349

rileyhCode:
- no longer use NLLanguageRecognizer which doesn't work for some languages (e.g., Latvian)
- don't specify a source language to the translator (use 'auto')
- no longer show the source language in the UI
- remove unused functions detectedLanguage


bnolens (approved):
LGTM

PR #217, opened 2023-12-21
MAM-3593-fix-an-issue-where-sometimes-users-re-experienced-onboarding

rileyhCode:
- fix the use of 'firstIndex' when looking for an account to update


bnolens (approved):
LGTM

PR #232, opened 2023-12-26
MAM-3611-add-support-for-proton-mail

rileyhCode:
- added support for Proton URL
- stop iterating over mail apps if there's a success
Annoyingly, Proton strips out the /n from the body (I even tried switching to \r to see if it made a difference; it didn't).

95DA61B6-2ED9-44F4-B8F5-0DC8C34A9512_1_102_o

### DIFF
--- a/Mammoth/Utils/EmailHandler.swift
+++ b/Mammoth/Utils/EmailHandler.swift
@@ -61,6 +61,7 @@ class EmailHandler: NSObject {
     }
 
 
+    static var emailAppURLIndex = 0
     fileprivate func openThirdPartyMailURLs(destinationEncoded: String, subjectEncoded: String, bodyEncoded: String) {
         // After lots of searching and digging, it appears there's a bug where calling
         // UIApplication.shared.canOpenURL() to check often fails, even when the correct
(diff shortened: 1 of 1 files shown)


Mammoth/Utils/EmailHandler.swift
```diff
@@ -72,18 +72,24 @@ class EmailHandler: NSObject {
...
-        }
-        if let sparkUrl = URL(string: sparkPrefix) {
-            UIApplication.shared.open(sparkUrl)
+        Task {
+            for prefixToTry in prefixesToTry {
+                if let urlToTry = URL(string: prefixToTry) {
+                    DispatchQueue.main.sync {
+                        UIApplication.shared.open(urlToTry) {success in
+                            if success {
+                                return
```
bnolens, line 84: The callback of `DispatchQueue.main.sync` is not returning anything. So I don't think you're aborting the for-loop early by returning here. What happens if multiple schemes are being opened?
You could use `await MainActor.run` instead which do return.

bnolens (changes requested):
Please review my comment.

Mammoth/Utils/EmailHandler.swift
rileyhCode: Good point. Made some changes; additional comments:

- have a static index into the Apps to try
- create a function to asynchronously try the next app in the list until there is a successful attempt
- added logging
- verified that if two of the apps are installed, only the first one in the list is notified

Mammoth/Utils/EmailHandler.swift
```diff
@@ -75,21 +76,35 @@ class EmailHandler: NSObject {
         let protonPrefix = "protonmail://mailto:\(destinationEncoded)?subject=\(subjectEncoded)&body=\(bodyEncoded)"
         let prefixesToTry = [gmailPrefix, outlookPrefix, yahooMailPrefix, sparkPrefix, protonPrefix]
         
-        Task {
-            for prefixToTry in prefixesToTry {
-                if let urlToTry = URL(string: prefixToTry) {
-                    DispatchQueue.main.sync {
```
bnolens: Another way would be to replace this line with `if let success = await MainActor.run { ...` and return if success is true.

bnolens (approved):
LGTM

Mammoth/Utils/EmailHandler.swift
rileyhCode: I wasn't quite able to get it to work this way, because UIApplication.shared.open() is an async call with a  completion routine. Because of that, there was no way to return a 'success' bool to MainActor.run.

PR #233, opened 2023-12-26
MAM-3592-app-icons-lag-when-loadingmisaligned-at-the-bottom

rileyhCode:
Fix from the internet…
Note: I was able to see this on the iPhone 15 Pro Max simulator, slowly scrolling and staring at the bottom row as it became visible.


bnolens (approved):
LGTM

PR #236, opened 2023-12-27
MAM-3608-ui-froze-after-deleting-a-post

bnolens:
https://linear.app/theblvd/issue/MAM-3608/ui-froze-after-deleting-a-post


rileyhCode (commented):
Do we know if this is still a problem on iOS < 16.0?

bnolens:
> Do we know if this is still a problem on iOS < 16.0?

Let me find a better fix that covers pre-16.0 versions too

rileyhCode (approved):
LGTM

PR #239, opened 2023-12-27
MAM-3615-crash-in-sidebarviewcontrollerdidmovetoparent

rileyhCode:
I was unable to reproduce the issue directly. Through code inspection, a nil superview will cause a crash, so I added a check for that.


bnolens (approved):
LGTM

PR #241, opened 2023-12-28
MAM-3587-add-to-list-context-menu-sometimes-broken-on-macos

rileyhCode:
- use UserCardModel.isFollowing to track if the user is following this account (it doesn't appear to be used for any other purpose)
- when creating the list of UserCardModels for the 'Following" list, set this to true
- include checking this flag when creating the contextual menu


bnolens (approved):
LGTM

PR #243, opened 2023-12-28
MAM-3617-post-actions-called-on-original-status-instead-of-local

bnolens:
[MAM-3617 : Post actions called on original status instead of local status](https://linear.app/theblvd/issue/MAM-3617/post-actions-called-on-original-status-instead-of-local-status)

Some of the post actions were using the "synced" status instead of the status from the user's instance. When we sync a post, we query the original post and update the PostCardModel with it's original status. We store the local status under preSyncedData. Use that status for post actions instead.


rileyhCode:
Interesting. I wonder if this is related to https://linear.app/theblvd/issue/MAM-3403/investigate-seeing-error-in-console-when-liking-a-post-from-a-hashtag

rileyhCode (approved):
LGTM. A good area for us to emphasize testing in.

PR #248, opened 2023-12-28
MAM-3620-sometimes-app-doesnt-remember-that-ive-onboarded

rileyhCode:
- use UserDefaults to track accounts that have onboarded
- added a one-time migration of acctData.wentThroughOnboarding
- mark wentThroughOnboarding as deprecated
- remove uses of wentThroughOnboarding


bnolens (approved):
LGTM!

PR #249, opened 2023-12-28
MAM-3621--ipados]-app-crash-when-sharing-post

rileyhCode:
iPads need a popoverPresentationController with a sourceView


bnolens (approved):
👍👍👍

PR #250, opened 2023-12-29
MAM-3616-app-freezes-when-mutingunmuting-accounts

bnolens:
[MAM-3616 : App freezes when muting/unmuting accounts](https://linear.app/theblvd/issue/MAM-3616/app-freezes-when-mutingunmuting-accounts)

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
@@ -375,7 +375,7 @@ extension NewsFeedViewController {
                     
                     cell.configure(postCard: model, type: viewModel.type.postCardCellType()) { [weak self] (type, isActive, data) in
                         guard let self else { return }
-                        guard !model.isDeleted else { return }
+                        guard !model.isDeleted, !model.isMuted, !model.isBlocked else { return }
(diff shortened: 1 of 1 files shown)


Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
```diff
@@ -718,6 +718,12 @@ extension NewsFeedViewController: NewsFeedViewModelDelegate {
                         CATransaction.commit()
                     }
                 }
+                
+                // This extra commit is needed when updating with .replaceAll (triggered by refresh snapshot)
+                // Without it the UI feezes.
+                if updateType == .replaceAll && updateDisplay {
+                    CATransaction.commit()
+                }
```
rileyhCode, line 726: It feels weird to have this in without an explicit matching CATransaction.begin(). (or maybe I'm missing it?)

rileyhCode (approved):
One comment, but not sure if there's anything to do about it.
Then, feel free to merge as you see fit.

Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
bnolens: Hate to admit but I have not yet found why the corresponding commit is not called in the `replaceAll` scenario. This is the only way I was able to fix it.

PR #251, opened 2023-12-29
MAM-3623-correctly-update-follow-status-on-profile-screen

bnolens:
[MAM-3623 : Correctly update follow status on Profile screen](https://linear.app/theblvd/issue/MAM-3623/correctly-update-follow-status-on-profile-screen)

When we don't have relationship information for a user we query that when opening a profile screen. The state of the follow button didn't update anymore to the correct state when the relationship object loaded.


rileyhCode (approved):
LGTM

PR #252, opened 2023-12-29
MAM-3624-stop-videos-in-quote-posts-when-cell-is-not-visible

bnolens:
[MAM-3624 : Stop videos in quote posts when cell is not visible](https://linear.app/theblvd/issue/MAM-3624/stop-videos-in-quote-posts-when-cell-is-not-visible)


rileyhCode (approved):
LGTM

PR #260, opened 2024-01-02
MAM-3632-on-ipad-tapping-mentions-doesnt-go-back-to-mentions-view

rileyhCode:
If the same button is pushed again, just pop other viewControllers.


bnolens (approved):
LGTM

PR #273, opened 2024-01-08
MAM-3678-sometimes-two-profileviewmodel-objects-are-being-created

rileyhCode:
Have the convenience init call the required init.

### DIFF
--- a/Mammoth/Screens/ProfileScreen/ProfileViewController.swift
+++ b/Mammoth/Screens/ProfileScreen/ProfileViewController.swift
@@ -54,8 +54,12 @@ class ProfileViewController: UIViewController, UIScrollViewDelegate, UITableView
 
     private var viewModel: ProfileViewModel
 
-    required init(user: UserCardModel? = nil, screenType: ProfileViewModel.ProfileScreenType = .own) {
-        self.viewModel = ProfileViewModel(.posts, user: user, screenType: screenType)
+    required init(user: UserCardModel? = nil, screenType: ProfileViewModel.ProfileScreenType = .own, viewModel: ProfileViewModel? = nil) {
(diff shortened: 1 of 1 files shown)


Mammoth/Screens/ProfileScreen/ProfileViewController.swift
```diff
@@ -54,8 +54,12 @@ class ProfileViewController: UIViewController, UIScrollViewDelegate, UITableView
 
     private var viewModel: ProfileViewModel
 
-    required init(user: UserCardModel? = nil, screenType: ProfileViewModel.ProfileScreenType = .own) {
-        self.viewModel = ProfileViewModel(.posts, user: user, screenType: screenType)
+    required init(user: UserCardModel? = nil, screenType: ProfileViewModel.ProfileScreenType = .own, viewModel: ProfileViewModel? = nil) {
```
bnolens, line 57: I'd propose a slightly different approach, one where we create an extra init function which only gets a viewModel argument. The problem with the current approach is that it requires you to understand how the method works to be able to call it correctly. If you pass both a `user` and a `viewModel` the user (and screenType) will be ignored.

bnolens (changes requested):
Proposing a slightly different approach. LMK what you think

Mammoth/Screens/ProfileScreen/ProfileViewController.swift
rileyhCode: Yeah; I get it. That makes sense.

rileyhCode:
- made the required init only take a viewModel
- made the original init a convenience init which calls the new required one

bnolens (approved):
LGTM!

PR #274, opened 2024-01-08
Mam 3669 not using remote instances for post fetching

rileyhCode:
- use the profile's instance to get their posts
- use the smart-list owner's instance to get their profile

NOTE: this is on top of MAM-3678

### DIFF
--- a/Mammoth/Screens/ProfileScreen/ProfileViewModel.swift
+++ b/Mammoth/Screens/ProfileScreen/ProfileViewModel.swift
@@ -293,11 +293,46 @@ extension ProfileViewModel {
 extension ProfileViewModel {
     private func createRequest(forType: ViewTypes, user: UserCardModel, range: RequestRange = .default) async throws -> (pinned: [Status], statuses: [Status], cursorId: String?) {
         do {
+            
+            // First, try to get posts directly from the profile's instance.
+            //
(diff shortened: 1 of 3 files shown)


Mammoth/Screens/ProfileScreen/ProfileViewModel.swift
```diff
@@ -293,11 +293,46 @@ extension ProfileViewModel {
 extension ProfileViewModel {
     private func createRequest(forType: ViewTypes, user: UserCardModel, range: RequestRange = .default) async throws -> (pinned: [Status], statuses: [Status], cursorId: String?) {
         do {
+            
+            // First, try to get posts directly from the profile's instance.
```
bnolens, line 297: This logic seems correct to me, but you should probably add it to `reloadUser` instead. That's where we already do this user lookup work when needed (there's probably a bug somewhere in `reloadUser` that needs to be solved). Also, reloadUser is called before `createRequest`, and only once, where `createRequest` is called on each page load since it's paginated content.

bnolens (changes requested):
Please review my comment

rileyhCode:
Closing this, and re-done in https://github.com/TheBLVD/mammoth/pull/277

PR #277, opened 2024-01-09
MAM-3669-v2-not-using-remote-instances-for-post-fetching

rileyhCode:
- use the profile's instance in ProfileViewModel.reloadUser()
- use the Smart List owner's domain/instance when possible


bnolens (approved):
LGTM

PR #278, opened 2024-01-09
MAM-3685-server-emoji-being-clipped-in-composer

rileyhCode:
- create EmojiCell (based on ImageCell)
- use scaleAspectFit for the emoji images
- remove extra cell setup calls from the viewController


bnolens (approved):
Nice!

PR #283, opened 2024-01-10
MAM-3703-usercardmodel-should-compare-emoji-to-be-equal

rileyhCode:
- have UserCardModel check .emojis for equality
- add equality to Emoji


bnolens (approved):
LGTM

PR #286, opened 2024-01-11
MAM-3708-retry-loading-feed-if-the-user-returns-a-500

rileyhCode:
- add a ClientError.networkError (with a status code) for network errors
- throw this error when necessary
- retry timeline requests for status code 206 and 500


bnolens (approved):
LGTM!

PR #289, opened 2024-01-12
MAM-3706-add-account-settings-option-in-account-switcher

bnolens:
[MAM-3706 : Add "Account settings" option in account switcher](https://linear.app/theblvd/issue/MAM-3706/add-account-settings-option-in-account-switcher)

### DIFF
--- a/Mammoth/Views/AccountSwitcherButton.swift
+++ b/Mammoth/Views/AccountSwitcherButton.swift
@@ -129,7 +129,7 @@ class AccountSwitcherButton: UIButton {
             allActions.append(accountAction)
         }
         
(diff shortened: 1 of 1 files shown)


rileyhCode (changes requested):
A couple notes:
1. the menu item should go to the Settings > Accounts pane (based on #design convo)
2. (not sure but…) I suspect pxl will want a different icon in that case

bnolens:
Forgot to push my changes. Will do now

rileyhCode (approved):
LGTM
One additional change for the icon to be circle-user per #design chat

PR #291, opened 2024-01-12
MAM-3716-fix-blank-labels-in-activity-cells

bnolens:
https://linear.app/theblvd/issue/MAM-3716/fix-blank-labels-in-activity-cells


rileyhCode (approved):
Reviewed. Added setting the postTextLabel attributes when the content is configured, so that content for Follows doesn't hilite content the same os other content.

PR #292, opened 2024-01-12
MAM-3661-user-infoavatar-wrong-in-for-you-for-a-few-seconds-in-feed

rileyhCode:
Only set the avatar image if self.user hasn't changed between the request and return of the image.


rileyhCode:
Self approving this one.

PR #294, opened 2024-01-14
updated mastodonkit to include search changes.

### LINKED ISSUE #287: unable to find a post/user by link.
### Describe the bug

when pasting a post's url to the search bar, it is expected for mastodon to find the post.
this isn't happening because the `resolve` parameter is not being set to true.

### To Reproduce

1. get a post url: `https://kolektiva.social/@piuvas/111733025336274898`.
2. paste it in the search bar in the discovery screen.
3. nothing.
4. press enter.
5. go to "posts".
6. also nothing.

### Expected behaviour

it shows the post both in regular AND posts search.

### Screenshots

image
image

### Instance address

kolektiva.social

### Device

iPhone 14 plus

### OS

iOS 17.2.1

### Version

build 474

### Additional context

_No response_

derspyy:
i couldn't tell if `resolve` should've been optional but i couldn't think of a situation in-app where it would've been false.

should close #287, question mark?

### DIFF
--- a/Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
+++ b/Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
@@ -330,8 +330,8 @@ public struct Accounts {
         let toLimitBounds = between(1, and: 80, default: 40)
         let parameters = [
             Parameter(name: "q", value: query),
-            Parameter(name: "resolve", true),
-            Parameter(name: "type", "accounts"),
+            Parameter(name: "resolve", value: true),
+            Parameter(name: "type", value: "accounts"),
             Parameter(name: "limit", value: limit.map(toLimitBounds).flatMap(toOptionalString)),
             Parameter(name: "following", value: following.flatMap(trueOrNil))
         ]
--- a/Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
+++ b/Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
@@ -330,12 +330,14 @@ public struct Accounts {
         let toLimitBounds = between(1, and: 80, default: 40)
         let parameters = [
             Parameter(name: "q", value: query),
+            Parameter(name: "resolve", true),
+            Parameter(name: "type", "accounts"),
             Parameter(name: "limit", value: limit.map(toLimitBounds).flatMap(toOptionalString)),
             Parameter(name: "following", value: following.flatMap(trueOrNil))
         ]
 
         let method = HTTPMethod.get(.parameters(parameters))
-        return Request<[Account]>(path: "/api/v1/accounts/search", method: method)
+        return Request<[Account]>(path: "/api/v2/search", method: method)
     }
     
     public static func registerAccount(username: String, email: String, password: String, agreement: Bool, locale: String) -> Request<LoginSettings> {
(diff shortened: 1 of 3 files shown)


derspyy:
i can also add a function like `searchPost()` to MastodonKit to allow us to get a single post from its URL and a `getPostFromUrl()` or something like that on `SearchService` (similar to the old `searchAccount()`), which would help with quote posts.

rileyhCode (changes requested):
I believe the goal of the PR is to fix #287 , correct? If so, I think we can just add this to Search.searchPosts():
`            Parameter(name: "resolve", value: "true")
`

derspyy:
well. the point is to match the mastodon api spec, that's only a consequence.

derspyy:
afaik, `v1/accounts/search` matches `@user@domain` and `v2/search` matches links.

derspyy:
(but for matching the username we are already using lookup)

rileyhCode:
The various search APIs are used in many disparate ways in the app (login instance search, Discover/search tab, looking up accounts, looking up posts, etc.). Because of this, I think it would be best if we can identify specific issues (like #287 above), and put up fixes for them. Ideally, these fixes will bring MastodonKit in line with the Mastodon APIs.

derspyy:
so you think it'd be better if i added those unused functions back in?

derspyy:
> I believe the goal of the PR is to fix #287 , correct? If so, I think we can just add this to Search.searchPosts():
> 
> `            Parameter(name: "resolve", value: "true")
> 
> `

i've renamed the issue to also address user urls, which is the other half of the pr.

derspyy:
the `Accounts.search()` url change was done because the spec says the old one's `resolve` param matches usernames, not urls. i've looked at the functions' usages throughout the code and didn't noticed anything that'd break. the only thing out of the blue was removing deprecated/unused functions which i think is a positive so we don't face this same problem in another interface later on.

derspyy:
anything else?

Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
```diff
@@ -330,12 +330,14 @@ public struct Accounts {
         let toLimitBounds = between(1, and: 80, default: 40)
```
rileyhCode, line 330: We should change this function name from search to searchAccounts so it better matches what it's doing.

Mammoth/Services/Backend/MastodonKit/Requests/Search.swift
```diff
@@ -18,36 +18,18 @@ public struct Search {
     public static func search(query: String, resolve: Bool? = nil) -> Request<Results> {
         let parameters = [
             Parameter(name: "q", value: query),
+            Parameter(name: "resolve", value: true),
```
rileyhCode, line 21: The resolve parameter is already added below. If it should always be true, then it shouldn't be a parameter to this function.
The way it's currently set up, the caller can specify if they want to resolve it. Maybe some of the call sites need updating?

rileyhCode (changes requested):

Mammoth/Services/Backend/MastodonKit/Requests/Accounts.swift
derspyy: will do!
although i have to add it's used as `Accounts.search()`.

Mammoth/Services/Backend/MastodonKit/Requests/Search.swift
derspyy: oh. got it.

derspyy: i just noticed that it's set to true in all calls. i wonder why it doesn't resolve to a profile on the broader search results. are we doing individual requests for these?
image

derspyy:
i thought of moving `searchAccounts` back to `Search.swift` since it's now using that endpoint. would that be less confusing?

PR #295, opened 2024-01-15
MAM-3660-support-multiple-image-sizes-in-post-cards

bnolens:
Support for multiple media sizes in PostCardCell and ActivityCardCell.
Media sizes: ~Dynamic~, Large, Small, Hidden
~Dynamic = when and images is < 200px width & height show it small, else show it large~
In small-mode, when a post has no post text, either show the text "(Image)", "(Video)", "(Gifv)" or "([type] description: [alt text])"

[MAM-3660 : Support multiple image sizes in post cards ✨](https://linear.app/theblvd/issue/MAM-3660/support-multiple-image-sizes-in-post-cards-✨)


Mammoth/Models/PostCardModel.swift
```diff
@@ -357,6 +357,15 @@ final class PostCardModel {
...
         self.hasMediaAttachment = self.mediaAttachments.count > 0
         
+        // The link to display
+        self.linkCard = status.reblog?.card ?? status.card
+        
+        // Post has a link to display
+        self.hasLink = self.linkCard?.url != nil
+        
+        // Hide the link image if there is a media attachment
+        self.hideLinkImage = true //self.hasMediaAttachment
```
rileyhCode, line 367: I *think* this is OK to leave as-it-is, but just confirming.

Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
```diff
@@ -111,20 +139,55 @@ final class PostCardCell: UITableViewCell {
...
+
+//                if mediaVariant == .auto {
+//                    if let firstImage = postCard.mediaAttachments.first,
+//                        let original = firstImage.meta?.original,
+//                        (original.width ?? 0) < 200 && (original.height ?? 0) < 200 {
+//                        mediaVariant = .small
+//                    } else {
+//                        mediaVariant = .large
+//                    }
+//                }
```
rileyhCode, line 161: Do you want to add a comment here (and maybe the same comment/text above as well) to easily find the areas to change if we use Dynamic mode?

rileyhCode (approved):
🌃 magic!
A couple notes to review, then feel free to merge.

rileyhCode:
I did see this blank video in /pixelart if you can take  a look:
image

rileyhCode:
Oh; it looks like that's a QT audio "movie", with an image you can see if you go into the detail view.
Maybe worth a separate ticket (?)

Mammoth/Models/PostCardModel.swift
bnolens: Yes, that's supposed to be set to `true` now. Keeping the `//self.hasMediaAttachment` adds a little context so if ok with you, I'd keep that comment for now as well.

rileyhCode:
Merging

PR #299, opened 2024-01-15
Add missing 'bookmarked' property to the CodingKeys of Status

nickludlam:
The `bookmarked` property was missing from CodingKeys, which means that no remotely bookmarked statuses would ever have their bookmarked property set true, preventing them from being correctly displayed, or capable of being removed within the app.

### DIFF
--- a/Mammoth/Services/Backend/MastodonKit/Models/Status.swift
+++ b/Mammoth/Services/Backend/MastodonKit/Models/Status.swift
@@ -103,6 +103,7 @@ public class Status: Codable, Hashable {
         case favouritesCount = "favourites_count"
         case reblogged
         case favourited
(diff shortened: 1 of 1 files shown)


rileyhCode:
Great catch.
Internal ticket: https://linear.app/theblvd/issue/MAM-3730/bookmarked-property-missing-from-codingkeys
(I also filed a ticket about the bookmarked posts not sorting as expected: https://linear.app/theblvd/issue/MAM-3729/bookmarked-posts-are-not-sorted-as-expected )

rileyhCode (approved):
👍

nickludlam:
Closing so that I can re-open it with a signed commit

PR #300, opened 2024-01-15
MAM-3451-post-removed-showing-for-multiple-people-in-feeds

bnolens:
When fetching a post returns a 404 (in feed sync logic), do a webfinger lookup and only delete the post if the account is federated.

[MAM-3451 : “Post removed” showing for multiple people in Feeds](https://linear.app/theblvd/issue/MAM-3451/post-removed-showing-for-multiple-people-in-feeds)

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
@@ -645,10 +645,15 @@ extension NewsFeedViewModel {
                 switch error as? ClientError {
                 case .mastodonError(let message):
                     if message == "Record not found" {
-                        if let postCard = item.extractPostCard() {
-                            let deletedPostCard = postCard
-                            deletedPostCard.isDeleted = true
-                            NotificationCenter.default.post(name: PostActions.didUpdatePostCardNotification, object: nil, userInfo: ["postCard": deletedPostCard])
+                        if let postCard = item.extractPostCard(), 
+                            let user = postCard.user,
+                            let instanceName = user.instanceName {
+                            // Do a webfinger lookup and only delete post if the account is federated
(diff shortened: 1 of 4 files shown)


rileyhCode (commented):
Interesting. I suspect this won't delete posts if the user's account is deleted though.
I wonder if that's something we need to consider. (?)
@jtomchak I wonder if there's a different check we can do to see if an account is federated?

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
```diff
@@ -645,10 +645,15 @@ extension NewsFeedViewModel {
...
                     if message == "Record not found" {
-                        if let postCard = item.extractPostCard() {
-                            let deletedPostCard = postCard
-                            deletedPostCard.isDeleted = true
-                            NotificationCenter.default.post(name: PostActions.didUpdatePostCardNotification, object: nil, userInfo: ["postCard": deletedPostCard])
+                        if let postCard = item.extractPostCard(), 
+                            let user = postCard.user,
+                            let instanceName = user.instanceName {
+                            // Do a webfinger lookup and only delete post if the account is federated
+                            if let webfinger = await AccountService.webfinger(user: user, serverName: instanceName), !webfinger.isEmpty {
```
bnolens, line 652: I'm wondering what the lookup returns if the account is deleted. Maybe not an empty string. Is there an account I can test this with?

rileyhCode: I don't know of one, but you can probably create one on moth.social, and then sign in and delete it (from Settings > Account Settings > Delete Account)

bnolens: I debugged it with posts of a deleted account and it worked

rileyhCode (approved):
Great; I think the additional check will be helpful here.

PR #301, opened 2024-01-15
MAM-3728-wrong-avatarimages-displayed-while-feed-loads

bnolens:
[MAM-3728 : Wrong Avatar/Images displayed while Feed loads](https://linear.app/theblvd/issue/MAM-3728/wrong-avatarimages-displayed-while-feed-loads)

### DIFF
--- a/Mammoth/Views/Cells/PostCardCell/PostCardProfilePic.swift
+++ b/Mammoth/Views/Cells/PostCardCell/PostCardProfilePic.swift
@@ -212,11 +212,11 @@ extension PostCardProfilePic {
         if let profileStr = self.user?.imageURL, let profileURL = URL(string: profileStr) {
             if self.profileImageView.sd_currentImageURL != profileURL {
                 self.profileImageView.sd_cancelCurrentImageLoad()
+                
+                self.profileImageView.ma_setImage(with: profileURL,
+                                                  cachedImage: self.user?.decodedProfilePic,
+                                                  imageTransformer: PostCardProfilePic.transformer) { image in }
             }
-            
-            self.profileImageView.ma_setImage(with: profileURL,
-                                              cachedImage: self.user?.decodedProfilePic,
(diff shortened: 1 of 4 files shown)


Mammoth/Views/Cells/PostCardCell/PostCardProfilePic.swift
```diff
@@ -168,6 +168,11 @@ extension PostCardProfilePic {
         
         if let profileStr = user.imageURL, let profileURL = URL(string: profileStr) {
             let userForImage = user
+            
+            if self.profileImageView.sd_currentImageURL != profileURL {
+                self.profileImageView.sd_cancelCurrentImageLoad()
+            }
```
rileyhCode, line 174: This will only cancel the current request if the (new) user.imageURL exists.
However, it won't cancel it if user.imageURL is nil. Is that a concern?

Mammoth/Views/Cells/PostCardCell/PostCardProfilePic.swift
```diff
@@ -202,6 +207,18 @@ extension PostCardProfilePic {
             self.onPress?(.profile, true, .user(user))
         }
     }
+    
+    func willDisplay() {
+        if let profileStr = self.user?.imageURL, let profileURL = URL(string: profileStr) {
+            if self.profileImageView.sd_currentImageURL != profileURL {
+                self.profileImageView.sd_cancelCurrentImageLoad()
```
rileyhCode, line 214: This will only cancel the current request if self.user?.imageURL exists.
However, it won't cancel it if is nil. Is that a concern?

rileyhCode (changes requested):
May need to call cancel in cases where the URL is nil

rileyhCode:
Merging after discussion.

rileyhCode (approved):
Discussed.

PR #302, opened 2024-01-16
Adding missing 'bookmarked' CodingKey to Status class

nickludlam:
The `bookmarked` property was missing from the coding keys in the `Status` class, which means that no remotely bookmarked statuses would ever have their bookmarked property set true, preventing them from being correctly displayed, or capable of being removed within the app.

This is the updated PR with a signed commit.


rileyhCode (approved):
LGTM
(sorry for the delay; was dealing with the CLA machinery)

PR #303, opened 2024-01-16
MAM-3707-when-login-into-custom-server-accept-https

bnolens:
[MAM-3707 : When login into custom server, accept https](https://linear.app/theblvd/issue/MAM-3707/when-login-into-custom-server-accept-https)

- strip https:// from instance name input when adding new account
When using https://moth.social the app returned an error. Trim the input into moth.social.


rileyhCode (approved):
LGTM

PR #311, opened 2024-01-18
MAM-3732-set-all-dependencies-to-fixed-versions

rileyhCode:
These were set to main/master. They have been changed as follows:
   MessageKit - 4.0.0+
   SDWebImage - 5.0.0+
   Kingfisher - 7.0.0+
   Reachability - a specific commit as it's version format doesn't work with Xcode (v5.1.0)

Before:
image

After:
image


bnolens (approved):
Awesome!

PR #316, opened 2024-01-18
(reapply) MAM-3752-multi-image-posts-in-activity-are-not-properly-defaulting-to…

rileyhCode:
Just reapplying the change for MAM-3752, as it depends on MAM-3757.


bnolens (approved):
LGTM :)

PR #321, opened 2024-01-19
MAM-3260-image-carousels-for-posts

bnolens:
[MAM-3260 : Image carousels for posts ✨](https://linear.app/theblvd/issue/MAM-3260/image-carousels-for-posts-✨)

MISSING:
- Missing a fixed aspect ratio lock when the original image or video is outside the 16/9 or 9/16 ratio
- In some cases videos are not displayed in their right aspect ratio
- accept swipes outside the gallery frame when the touch is on content of the scrollview (overriding hitTest)

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
@@ -699,7 +699,7 @@ extension NewsFeedViewController: NewsFeedViewModelDelegate {
         
         let updateDisplay = (NewsFeedTypes.allActivityTypes + [.mentionsIn, .mentionsOut]).contains(feedType) || self.isInWindowHierarchy()
         
-        guard !self.tableView.isDragging, updateDisplay else {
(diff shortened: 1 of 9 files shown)


Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
```diff
@@ -699,7 +699,7 @@ extension NewsFeedViewController: NewsFeedViewModelDelegate {
         
         let updateDisplay = (NewsFeedTypes.allActivityTypes + [.mentionsIn, .mentionsOut]).contains(feedType) || self.isInWindowHierarchy()
         
-        guard !self.tableView.isDragging, updateDisplay else {
+        guard !self.tableView.isTracking, updateDisplay else {
```
bnolens, line 702: I didnt' meant to include this. It is an improvement though. `isTracking` is covering the case where the user is dragging but is not moving as well, which we need to do.

rileyhCode (approved):
Awesome.

PR #324, opened 2024-01-19
MAM-3713-fix-issue-with-bookmark-fetching-threshold

rileyhCode:
Update bookmarks even if there's only one new item.


bnolens (approved):
LGTM

PR #325, opened 2024-01-19
MAM-3352-links-on-profiles-not-working-build-340

rileyhCode:
When switching accounts, ProfileHeader.onUserActionPress was nil (and so, tapping the field resulted in nothing happening). Now, onUserActionPress is set when ProfileHeader is created.

Also: clear self.extraInfoConstraints once it's associated data has been cleared.


bnolens (approved):
LGTM

PR #332, opened 2024-01-22
MAM-3724-image-sizecropping-in-mammoth

bnolens:
[MAM-3724 : Image Size/cropping in Mammoth](https://linear.app/theblvd/issue/MAM-3724/image-sizecropping-in-mammoth)

### DIFF
--- a/Mammoth/Services/Backend/MastodonKit/Models/Attachment.swift
+++ b/Mammoth/Services/Backend/MastodonKit/Models/Attachment.swift
@@ -98,7 +98,7 @@ public class AttachmentMeta2: Codable {
     init(width: Int, height: Int) {
         self.width = width
         self.height = height
-        self.aspect = Double(max(Float(width), 1.0) / Float(height))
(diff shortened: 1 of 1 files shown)


Mammoth/Services/Backend/MastodonKit/Models/Attachment.swift
```diff
@@ -98,7 +98,7 @@ public class AttachmentMeta2: Codable {
     init(width: Int, height: Int) {
         self.width = width
         self.height = height
-        self.aspect = Double(height / max(width, 1))
+        self.aspect = Double(max(Float(width), 1.0) / Float(height))
```
rileyhCode, line 101: The max() is to prevent a divide by zero, so should apply to the height, not the width.

rileyhCode (changes requested):
Good catch 👀. One note on the max() call.

rileyhCode (approved):
LGTM

PR #335, opened 2024-01-22
MAM-3777-enabling-notifications-on-a-profile-returns-a-404

bnolens:
[MAM-3777 : Enabling notifications on a profile returns a 404](https://linear.app/theblvd/issue/MAM-3777/enabling-notifications-on-a-profile-returns-a-404)

### DIFF
--- a/Mammoth/Utils/UserActions.swift
+++ b/Mammoth/Utils/UserActions.swift
@@ -132,7 +132,7 @@ struct UserActions {
         let vc = NewPostViewController()
         vc.isModalInPresentation = true
         vc.fromPro = true
-        vc.proText = "@\(user.acct) "
+        vc.proText = "\(user.remoteFullOriginalAcct) "
         target.present(UINavigationController(rootViewController: vc), animated: true)
     }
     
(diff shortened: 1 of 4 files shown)


Mammoth/Utils/UserActions.swift
```diff
@@ -132,15 +132,15 @@ struct UserActions {
         let vc = NewPostViewController()
         vc.isModalInPresentation = true
         vc.fromPro = true
-        vc.proText = "@\(user.acct) "
```
rileyhCode, line 135: In general, the .acct would be preferred as it's just the @username if the mentioned account is on the same instance as the signed in user. (but of course, being correct is even more important). Was this broken?

rileyhCode (commented):
I played with this, and it did seem to fix the case outlined in the ticket.

Mammoth/Utils/UserActions.swift
bnolens: I believe this is broken. These are my steps:
1. I'm logged in as bnolens@staging.moth.social
2. I search for bnolens@moth.social on the discovery tab
3. I open that user's profile and hit "Mention" in the top right settings menu
4. The compose is prefilled with "@bnolens". 
5. I send that message and navigate to my profile
6. Tapping on @bnolens inside the new message navigates to bnolens@staging.moth.social and not bnolens@moth.social

rileyhCode (approved):
LGTM then.

PR #339, opened 2024-01-23
MAM-3787-fix-profile-pagination-errors-after-user-sync

bnolens:
When visiting someone's profile we fetch the user object from its original instance and replace the user object (UserCardModel) in ProfileViewModel with it. Later, we recreated the user instead of mutating it to update the user but forgot to copy over the instanceName. This lead to a broken pagination on profile screen when the original user was from a different instance as the logged in instance. 

[MAM-3787 : Fix profile pagination errors after user sync](https://linear.app/theblvd/issue/MAM-3787/fix-profile-pagination-errors-after-user-sync)


rileyhCode (approved):
Great catch/fix.

PR #344, opened 2024-01-23
MAM-3788-prevent-infinite-loop-when-fetchfollowingtags-fails

bnolens:
[MAM-3788 : Prevent infinite loop when fetchFollowingTags fails](https://linear.app/theblvd/issue/MAM-3788/prevent-infinite-loop-when-fetchfollowingtags-fails)


rileyhCode (approved):
LGTM

PR #349, opened 2024-01-23
MAM-3794-media-helper-text-in-small-mode-only-for-status

bnolens:
[MAM-3794 : Media helper text in small-mode only for status notifications](https://linear.app/theblvd/issue/MAM-3794/media-helper-text-in-small-mode-only-for-status-notifications)

It should only be saying (Carousel) in New Post notifications (in small-media mode)
IMG_4A6D5BF29E08-1


rileyhCode (approved):
LGTM

PR #353, opened 2024-01-24
MAM-3800-image-carousels-misalign-when-exiting

bnolens:
[MAM-3800 : Image carousels misalign when exiting](https://linear.app/theblvd/issue/MAM-3800/image-carousels-misalign-when-exiting)


rileyhCode (approved):
LGTM

PR #363, opened 2024-01-26
MAM-3817-fix-activitycardcell-repeated-media-bug

bnolens:
Create media components once in setupUI and use isHidden to show/hide when needed. This should prevent duplicate media creation. This brings it also more inline with the architecture of the PostCardCell.

[MAM-3817 : Fix ActivityCardCell repeated media bug](https://linear.app/theblvd/issue/MAM-3817/fix-activitycardcell-repeated-media-bug)


jtomchak (approved):
oh much better

PR #366, opened 2024-01-27
MAM-3823-mammoth-sends-the-user-token-to-unknown-servers

rileyhCode:
Only include the token when communicating to the user's instance.
Allow for the mothClient to use its token.
Allow certain URLs to use their tokens.


rileyhCode:
Merging.

PR #369, opened 2024-01-29
MAM-3829-switching-profiles-from-compose-screen-doesnt-change-the

bnolens:
Use the custom emojis of the server of the current account in the composer, instead of the emojis of the current account in the rest of the app.

[MAM-3829 : Switching profiles from compose screen doesn't change the custom emoji picker.](https://linear.app/theblvd/issue/MAM-3829/switching-profiles-from-compose-screen-doesnt-change-the-custom-emoji)


rileyhCode (approved):
LGTM

PR #370, opened 2024-01-29
MAM-3834-unsubscribe-subscribe-blocking-client

bnolens:
Don't call `/for_you/me` right after subscribe/unsubscribe of smart lists, but use the subscribe/unsubscribe network call result to update the local user. 

[MAM-3834 : Unsubscribe / Subscribe blocking (client)](https://linear.app/theblvd/issue/MAM-3834/unsubscribe-subscribe-blocking-client)


rileyhCode (approved):
LGTM

PR #372, opened 2024-01-29
MAM-3811-cannot-upload-image-video

rileyhCode:
- put up an alert if the user tries to upload more than one video, or a combination of video + image
- generalize the media error alert


bnolens (approved):
LGTM

PR #378, opened 2024-02-01
Added two expected ArkanaKeys env entries in the sample

nickludlam:
This is a short commit to bring the sample ArkanaKeys env file up to date with what the `.arkana.yml` expects.


nickludlam:
Actually I've just noticed this also contains the previous commit I made with Bookmark CodingKeys. I will re-file

PR #380, opened 2024-02-01
Prevent the bookmark property from being incorrectly overwritten

nickludlam:
Currently the `PostCardModel.mergeInOriginalData()` method (used for syncing accurate like, repost and reply counts) will cause the bookmarked property to be overwritten. Since bookmark status is only accurate when you're querying your home Mastodon server, this property is always being trampled.

This PR simply makes the bookmarks property unable to be made false by a merge, if it was previously true.

### DIFF
--- a/Mammoth/Models/PostCardModel.swift
+++ b/Mammoth/Models/PostCardModel.swift
@@ -668,7 +668,13 @@ final class PostCardModel {
         if self.preSyncData == nil {
             self.preSyncData = self.data
         }
-        // updating the data object so that computed properties 
+        
+        // bookmarked status is local authoritative only, so remote fetched data will always be wrong
+        if (self.isBookmarked) {
+            newStatus.bookmarked = true
+        }
+
+        // updating the data object so that computed properties
         // e.g. likesCount is getting updated
         if self.isReblogged {
             if case .mastodon(let status) = data {
--- a/Mammoth/Models/PostCardModel.swift
+++ b/Mammoth/Models/PostCardModel.swift
@@ -673,6 +673,11 @@ final class PostCardModel {
         if (self.isBookmarked) {
             newStatus.bookmarked = true
         }
+        
+        // Same with if a toot has been liked outside the app, and hasLocalMetric() cannot override it
+        if (self.isLiked) {
+            newStatus.favourited = true
+        }
 
         // updating the data object so that computed properties
         // e.g. likesCount is getting updated


rileyhCode (commented):
I wonder if it's better to change .isBookmarked to check .preSyncData first, and fallback to .data.
@bnolens for comment.

derspyy:
i feel like this bookmark data should always be ignored, since it's always wrong.

nickludlam:
Just to add to this, if someone has favorited/liked a toot outside of Mammoth, it also gets trampled.

@derspyy Unfortunately the remote server response is coming back with a definite 'bookmarked: false' property, so there is no null/undefined state to allow us to choose not to merge that property. The only other alternative would be to track which server this data was pulled from, and could be compared with `AccountsManager.shared.currentAccountClient.baseHost`.

derspyy:
> Just to add to this, if someone has favorited/liked a toot outside of Mammoth, it also gets trampled.
> 
> @derspyy Unfortunately the remote server response is coming back with a definite 'bookmarked: false' property, so there is no null/undefined state to allow us to choose not to merge that property. The only other alternative would be to track which server this data was pulled from, and could be compared with `AccountsManager.shared.currentAccountClient.baseHost`.

i was under the assumption that even if the post is from our instance we wouldn't be calling this at all.

bnolens:
For me the right approach here is to define a list of keys we want to merge into the current status, and ignore all the rest. At the moment [we replace the entire status](https://github.com/TheBLVD/mammoth/blob/d09b82ec2477648ea099da916e7454c2e8edf28c/Mammoth/Models/PostCardModel.swift#L679). Keys like `favourited` and `reblogged` we should ignore and only merge in `favouritesCount`, `reblogsCount`, and others we're interested in.

derspyy:
> For me the right approach here is to define a list of keys we want to merge into the current status, and ignore all the rest. At the moment [we replace the entire status](https://github.com/TheBLVD/mammoth/blob/d09b82ec2477648ea099da916e7454c2e8edf28c/Mammoth/Models/PostCardModel.swift#L679). I would stop doing that. Keys like `favourited` and `reblogged` we should ignore and only merge in `favouritesCount`, `reblogsCount`, and others we're interested in.

do you think this should be a ticket?

bnolens (approved):
I think this is a good patch for now. Hoping to be able to apply the architectural changes explained above in a second phase. (only replace explicit keys instead of the entire status object)

PR #383, opened 2024-02-01
MAM-3862-high-contrast-issue-with-some-mentions-content

rileyhCode:
- pass in postCard data when configuring replyButton and likeButton
- use postCard.isPrivateMention when setting background colors
- track isPrivateMention in PostCardProfilePic as well


bnolens (approved):
LGTM

PR #384, opened 2024-02-02
MAM-3841-integrate-react-native

bnolens:
- Integrated React Native
- Added a simple composer UI using RN + native components
- Updated README


README.md
```diff
@@ -12,18 +12,35 @@ The Mammoth Team
 **Requirements:**
 
 - Ruby 3.1+
-- Node.js 16+
+- Node.js 20.10.+
 - Xcode 13+
 - Swift 5+
+- Yarn 1.22+
```
rileyhCode, line 18: Should we make a callout that this isn't the latest version?
(maybe point people here? not sure https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable)

README.md
```diff
@@ -12,18 +12,35 @@ The Mammoth Team
...
 
 1. Clone the repo
 1. Copy the `sample.env` to `.env`
-1. Run `bundle install` to get the needed gems
-1. From the root of the application still run `bin/arkana` This will generated
+1. Run `yarn` to install dependencies
```
rileyhCode, line 22: What directory should the user be in?

README.md
```diff
@@ -12,18 +12,35 @@ The Mammoth Team
...
+1. From the [ios](./ios) directory run `bundle install` to get the needed gems
+1. From the [ios](./ios) directory run `bin/arkana` This will generated
```
rileyhCode, line 24: Note: I had to move the .arkana.yml into the ios directory.

ios/.xcode.env.local
```diff
@@ -0,0 +1,2 @@
+export NODE_BINARY=/opt/homebrew/bin/node
```
rileyhCode, line 1: Mine was in a different location. Maybe this should be `which node` or…?

rileyhCode (commented):
I added some thoughts.

README.md
bnolens: Is this not the latest version? I'm not trying to run an older version here. To me this looks like the latest 
Screenshot 2024-02-05 at 10 38 55

bnolens: I didn't meant to run an older version. Good catch! I updated it to the latest version.

ios/.xcode.env.local
bnolens: This file should be ignored by git and generated when running `... pod install`

bnolens: I removed it from git. So next time you run `USE_FRAMEWORKS=static RCT_NEW_ARCH_ENABLED=1 arch -x86_64 bundle exec pod install` the file should be generated locally

.yarnrc.yml
```diff
@@ -0,0 +1,3 @@
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.1.0.cjs
```
jtomchak, line 3: creates errors on fresh installs, remove `yarnPath:`

> Now that Corepack is shipped with both Node 18 and 20 we no longer need to rely on [yarnPath](https://yarnpkg.com/configuration/yarnrc#yarnPath), and as a result we updated our [installation guide](https://yarnpkg.com/getting-started/install) to reflect that. The yarn [init](https://yarnpkg.com/cli/init) -2 and yarn set version commands have been updated to favor updating the [packageManager](https://yarnpkg.com/configuration/manifest#packageManager) field when possible.

jtomchak (approved):
looks great

PR #385, opened 2024-02-02
MAM-3870-yellow-highlights-missing-from-the-discover-tab

rileyhCode:
Reload string content on theme change since they have colored content.


bnolens (approved):
LGTM

PR #391, opened 2024-02-06
MAM-3889-like-actions-not-sticking

bnolens:
[MAM-3889 : like actions not sticking](https://linear.app/theblvd/issue/MAM-3889/like-actions-not-sticking)
[MAM-3894 : Like not working on post details](https://linear.app/theblvd/issue/MAM-3894/like-not-working-on-post-details)

### DIFF
--- a/Mammoth/Views/Cells/PostCardCell/PostCardFooter.swift
+++ b/Mammoth/Views/Cells/PostCardCell/PostCardFooter.swift
@@ -209,7 +209,7 @@ extension PostFooterButton {
     func configure(buttonText: String?, isActive: Bool = false, postCard: PostCardModel? = nil) {
         self.isActive = isActive
         
-        let shouldChangeTheme = self.isPrivateMention != postCard?.isPrivateMention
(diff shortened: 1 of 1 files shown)


Mammoth/Views/Cells/PostCardCell/PostCardFooter.swift
```diff
@@ -201,6 +208,8 @@ fileprivate class PostFooterButton: UIButton {
 extension PostFooterButton {
     func configure(buttonText: String?, isActive: Bool = false, postCard: PostCardModel? = nil) {
         self.isActive = isActive
+        
+        let shouldChangeTheme = self.isPrivateMention != postCard?.isPrivateMention
```
rileyhCode, line 212: I think this line should be testing the same value as the line below:
`let shouldChangeTheme = self.isPrivateMention != (postCard?.isPrivateMention ?? false)`

rileyhCode (approved):
Good fix. Pls review comment, then merge as you see fit.

PR #395, opened 2024-02-06
MAM-3896-performance-do-not-update-the-feed-when-scrolling-and

bnolens:
When new content is being injected in the feed when scrolling, an animation hitch could occur. To prevent this we might want to only update the feed when it's not moving. 

IMPORTANT to do an in-depth QA of the feed experience before we ship this.

[MAM-3896 : Performance: do not update the feed when scrolling and decelerating](https://linear.app/theblvd/issue/MAM-3896/performance-do-not-update-the-feed-when-scrolling-and-decelerating)


rileyhCode (approved):
LGTM. At first glance, it seems quite noticeable.

PR #396, opened 2024-02-06
MAM-3881-likesrepliesreposts-line-should-be-same-color-as-view-details

rileyhCode:
Set likes/replies/reposts color to match the buttons


bnolens (approved):
LGTM!

PR #399, opened 2024-02-07
Fix bookmark pagination

nickludlam:
This PR fixes bookmark pagination, so you can scroll back through your list and have it fetch the paginated requests correctly.

It has required splitting the bookmark fetch calls out into a paginated request, then following that change up through the layers into `NewsFeedViewModel`.

More information outlined here https://linear.app/theblvd/issue/MAM-3869/bookmark-pagination-is-broken


nickludlam:
Closing this as it's not quite right at the moment, I've missed some elements about how the cursorId value in `NewsFeedViewModel` works!

PR #401, opened 2024-02-07
Add clearCache() to StatusCache and hook up to clearAllCaches()

nickludlam:
During my bookmark testing I noticed we did not purge any of the state accumulated in `StatusCache`, so this adds a basic method and hooks it up to `SettingsViewController.clearAllCaches()`.


bnolens (approved):
LGTM

PR #408, opened 2024-02-09
MAM-3910-post-card-read-more-button-on-text-truncation

bnolens:
[MAM-3910 : Post card read more button on text truncation](https://linear.app/theblvd/issue/MAM-3910/post-card-read-more-button-on-text-truncation)

https://github.com/TheBLVD/mammoth/assets/221925/1934729d-47d3-4f44-bd92-f97d438799b8

### DIFF
--- a/Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
+++ b/Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
@@ -951,7 +951,8 @@ extension PostCardCell {
             }
             
             self.readMoreButton?.isHidden = true
-            DispatchQueue.main.async {
+            DispatchQueue.main.async { [weak self] in
+                guard let self else { return }
                 if self.postTextView.isTruncated && self.type != .detail {
                     self.readMoreButton?.isHidden = false
                     self.postTextView.bringSubviewToFront(self.readMoreButton!)
--- a/Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
+++ b/Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
@@ -350,6 +350,8 @@ final class PostCardCell: UITableViewCell {
     private var headerExtension: PostCardHeaderExtension?
     private var metadata: PostCardMetadata?
     
+    private var readMoreButton: ReadMoreButton?
+    
     private var cellVariant: PostCardVariant {
         if let reuseIdentifier = self.reuseIdentifier {
             return Self.variant(for: reuseIdentifier)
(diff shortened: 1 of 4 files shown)


Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
```diff
@@ -934,6 +949,16 @@ extension PostCardCell {
                     self.postTextView.isHidden = true
                 }
             }
+            
+            self.readMoreButton?.isHidden = true
+            DispatchQueue.main.async {
```
rileyhCode, line 954: I'm a little concerned about the case where configure() is called on this same cell (for different content) while this dispatch is still pending. Can it be cancelled/ignored if the postCardModel changes?

rileyhCode (changes requested):
One question about the dispatch to update the button.

Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
bnolens: If configure is called again, this code will be called again too. Isn't that what we want? @rileyhCode 
This makes me think. Maybe this should be implemented in MetaLabel. The reason why this code needs to run on the next runloop is because a new layout is requested when the label is configured. Only after that layout pass can we say what the height of the content is. 
Maybe a refactoring I can do when we decide this ReadMore approach is the one we want?

rileyhCode: It's correct that this will get re-called when configure() is re-called. So, it will eventually be correct; however, it could flash to the incorrect value.

For example: config A needs 'more…', config B does not:

- prepareForReuse() - button hidden
- configure(A)
- prepareForReuse() - button hidden
- configure(B)
- dispatchMain(A) - button shows
- dispatchMain(B) - button hidden

Having said that… maybe it's not an issue in practice, and something we can just be aware of, and keep an eye out for.

bnolens: You're right it's not a good pattern. Thinking how I could improve it..
I haven't seen any flickering yet in practise (mainly because cells are configured before appearing on screen), but I agree with could technically happen.

bnolens: To me the right solution is to redo this implementation in the MetaLabel library and set the .isHidden value inside the `layoutSubviews` method.

rileyhCode (approved):
LGTM

PR #409, opened 2024-02-09
MAM-3907-repetition-bug-in-quote-posts

bnolens:
Refactor how we configure the quote post UI. Instead of creating it in`configure(...)` (in realtime) we create all elements once in `setupUI()` and use .isHidden to enable what we need in `configure()`. This approach is what we started doing in all major cell components.

[MAM-3907 : Repetition bug in quote posts](https://linear.app/theblvd/issue/MAM-3907/repetition-bug-in-quote-posts)

### DIFF
--- a/Mammoth/Screens/DiscoveryScreen/PostResultsViewModel.swift
+++ b/Mammoth/Screens/DiscoveryScreen/PostResultsViewModel.swift
@@ -227,12 +227,12 @@ private extension PostResultsViewModel {
                         // Delete post card data in list data
                         self.listData.remove(at: cardIndex)
                         // Request a table view cell refresh
-                        self.delegate?.didDeleteCard(at: IndexPath(row: cardIndex, section: 1))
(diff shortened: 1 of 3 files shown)


Mammoth/Screens/DiscoveryScreen/PostResultsViewModel.swift
```diff
@@ -227,12 +227,12 @@ private extension PostResultsViewModel {
                         // Delete post card data in list data
                         self.listData.remove(at: cardIndex)
                         // Request a table view cell refresh
-                        self.delegate?.didDeleteCard(at: IndexPath(row: cardIndex, section: 1))
+                        self.delegate?.didDeleteCard(at: IndexPath(row: cardIndex, section: 0))
```
rileyhCode, line 230: Great catch!

rileyhCode (approved):
LGTM

PR #411, opened 2024-02-09
MAM-3928-use-new-follow-suggestions-endpoint

rileyhCode:
- stop requesting follow suggestions in DiscoverViewModel (the model used for showing Account search results; it doesn't need the suggestions)
- create a featureClient, similar to client, mothClient
- use the featureClient to get follow suggestions


jtomchak (approved):
Looks good. Let’s give it a spin

PR #412, opened 2024-02-09
MAM-3924-include-is-gold-member-status-in-the-daily-checkin

rileyhCode:
- add gold status to JWT
- call activities() when gold status changes, and the user is a gold member


bnolens (approved):
LGTM, wrote some thoughts on Discord about our approach though

PR #424, opened 2024-02-15
MAM-3951-macos-no-suggested-follows

rileyhCode:
Reinstate the initial request for account content. It's not needed for iPhone, but is for iPad/macOS.


bnolens (approved):
LGTM

PR #426, opened 2024-02-19
MAM-3963-animated-gifs-display-incorrectly-in-account-switcher-button

rileyhCode:
Set the UIButton.imageView layer cornerRadius instead of using SDImagePipelineTransformer which doesn't seem to work for GIFs on UIButtons.
(It does seem to work on UIImageViews as in PostCardProfilePic)


bnolens (approved):
The way I solved this in the PostCardProfilePic is to keep both transformer and cornerRadius. Assuming the masking when using cornerRadius is less expensive if there's nothing to mask. I have no data to prove this though. In this PRs case, performance is less of an issue. So this looks good to me.

PR #427, opened 2024-02-22
MAM-3642-add-a-bottom-up-feed-behavior-as-in-ivory

bnolens:
Changes:

- [ ] A new option in the settings to enable top-down or bottom-up behavior  (*NOT DONE YET*)
- [x] Set default behavior to be bottom-up
- [x] An unread indicator (we already have this)
- [x] A secondary load mechanism. 
- [x] Instead of loading the 60 newest posts + load more button (if needed) we load all newest posts, starting with the next 20 newer posts. 

Ticket with more context: 
https://linear.app/theblvd/issue/MAM-3642/add-a-bottom-up-feed-behavior-as-in-ivory-%E2%9C%A8

changes also inspired by work by @derspyy on https://github.com/TheBLVD/mammoth/tree/timeline-completion

Context

Some of our users prefer the feed behavior of Ivory (bottom-up instead of top-down); 

Ivory inserts new posts at the top of the feed in a strictly chronological order. The feed will periodically load all new posts and insert them at the top of the feed when the app is foreground. They never show a Load More button or any type of gap in the feed. 

Instead of showing a face pill, in Ivory they display an unread indicator with the amount of unread posts higher in the feed. The indicator is decreasing while the user is scrolling up. 

In Ivory, the feed is loading new posts in chunks of 200 posts (testing on the moth.social Federated feed). It will recursively load these chunks in until all new posts are loaded and inserted in the feed. The unread indicator is reflecting this behavior. When 840 new new posts are available, the indicator will increase with increments of max 200 until reaching the final number of 840.


derspyy:
my observations: 
- i feel like we should fetch posts in batches of 40, instead of 20, so that we don't hit a rate limit.
- are we automatically downloading the media of theses posts? we should look into only fetching new media (not cached) when the post is scrolled to view, so that we don't download images off of posts that are scrolled past if the user is scrolling quickly.

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
```diff
@@ -59,13 +65,6 @@ extension NewsFeedViewModel {
                             self.state = .success
                             self.hideLoader(forType: currentType)
                             
-                            // Preload quote posts
-                            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
-                                uniqueNewItems.forEach({
-                                    $0.extractPostCard()?.preloadQuotePost()
-                               })
-                            }
-
```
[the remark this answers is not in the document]
bnolens: we do this in the `willDisplay` method of the PostCardCell now, which is called when the cell is about the be displayed

bnolens:
https://github.com/TheBLVD/mammoth/assets/221925/4fb6a70b-2411-473e-b21b-f5d74c0d613b



> my observations:
> 
> * i feel like we should fetch posts in batches of 40, instead of 20, so that we don't hit a rate limit.
> * are we automatically downloading the media of theses posts? we should look into only fetching new media (not cached) when the post is scrolled to view, so that we don't download images off of posts that are scrolled past if the user is scrolling quickly.

I pushed an update with the batch limit on 40. Seems to work fine (see video here)
The media is only loaded with the cell is about to be displayed. See the method `preloadCards(atIndexPaths indexPaths: [IndexPath])` to find this code.

https://github.com/TheBLVD/mammoth/assets/221925/46a5113a-7c86-4699-b05a-1b392cae8466

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
```diff
@@ -713,7 +737,6 @@ extension NewsFeedViewModel {
                         }
                     }
                 default:
-                    NotificationCenter.default.post(name: PostActions.didUpdatePostCardNotification, object: nil, userInfo: ["postCard": postCard])
```
rileyhCode, line 716: I wonder what case this was handling (if any).

Mammoth/Views/Cells/PostCardCell/PostCardProfilePic.swift
```diff
@@ -171,18 +171,6 @@ extension PostCardProfilePic {
...
-            
-            self.profileImageView.ma_setImage(with: profileURL,
-                                              cachedImage: self.user?.decodedProfilePic,
-                                              imageTransformer: PostCardProfilePic.transformer) { image in
-                if userForImage == self.user {
-                    user.decodedProfilePic = image
-                }
-            }
-        }
-
```
rileyhCode, line 185: How is the profile image set, if not here?

rileyhCode: Oh; I see. In willDisplay().

rileyhCode (approved):
Overall, LGTM. I am a bit concerned that the switches to MainActor.run in a few places may have unintended consequences, but I don't have a concrete example.

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
bnolens: I just can't see a reason for this notification post to be there. And having it post these notifications unintentionally could have really bad consequences for scroll performance. So I believe it's better to remove than to keep.

PR #428, opened 2024-02-22
Only run filters and blocks client-side on remote feeds.

derspyy:
this choice was made because on authenticated requests, these are handled by the server.

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
@@ -44,6 +44,8 @@ extension NewsFeedViewModel {
                         newItems = items.removeMutesAndBlocks().removeFiltered()
                     } else if case .forYou = currentType {
                         newItems = items.removeMutesAndBlocks().removeFiltered()
+                    } else if case .channel = currentType {
+                        newItems = items.removeMutesAndBlocks().removeFiltered()
                     } else {
                         newItems = items
                     }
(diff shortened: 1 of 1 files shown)


Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
```diff
@@ -37,8 +37,16 @@ extension NewsFeedViewModel {
 
                 if let lastId = self.oldestItemId(forType: currentType) {
                     let (items, cursorId) = try await currentType.fetchAll(range: RequestRange.max(id: lastId, limit: 20), batchName: "next-page_batch")
-  
-                    let newItems = items.removeMutesAndBlocks().removeFiltered()
+                    
+                    // only remove mutes and blocks in remote feeds.
+                    let newItems: [NewsFeedListItem]
+                    if case .community = type {
```
bnolens, line 43: Also for `.channel(let type)`, right?

bnolens (changes requested):

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
derspyy: oh! are those... smart lists?

bnolens: Yes

bnolens: We called them channels before launch

bnolens (approved):
LGTM!

PR #432, opened 2024-02-26
MAM-3979-migrate-strings-to-string-catalogs

derspyy:


Mammoth/Localizable.xcstrings
```diff
@@ -0,0 +1,639 @@
...
+          }
+        }
+      }
+    },
+    "title.trending" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Trensing"
```
rileyhCode, line 629: Trensing -> Trending

rileyhCode (commented):
Just a general note that the strings that depend on a count (e.g., post.likeCount) need a bit more work. There is some machinery in iOS to make it pick the correct value based on the # of items.

There's a more complete description of the issue here (although… I think the solution they present is a bit outdated, meaning Apple has made a more streamlined solution available): https://medium.com/@vitaliikuznetsov/plurals-localization-using-stringsdict-in-ios-a910aab8c28c

derspyy:
> Just a general note that the strings that depend on a count (e.g., post.likeCount) need a bit more work. There is some machinery in iOS to make it pick the correct value based on the # of items.
> 
> 
> 
> There's a more complete description of the issue here (although… I think the solution they present is a bit outdated, meaning Apple has made a more streamlined solution available): https://medium.com/@vitaliikuznetsov/plurals-localization-using-stringsdict-in-ios-a910aab8c28c

i'm currently using plurals internally. maybe they're not immediately visible as text, but on xcode, it's there.

Mammoth/Views/Cells/PostCardCell/PostCardMetadata.swift
```diff
@@ -135,9 +135,12 @@ final class PostCardMetadata: UIView {
             }
         }
         
-        likesLabel.text = Int(postCard.likeCount) == 1 ? "1 Like" : "\(postCard.likeCount) Likes"
-        repliesLabel.text = Int(postCard.replyCount) == 1 ? "1 Reply" : "\(postCard.replyCount) Replies"
-        repostsLabel.text = Int(postCard.repostCount) == 1 ? "1 Repost" : "\(postCard.repostCount) Reposts"
+        let like_localized = postCard.likeCount == "1" ? "post.likeCount.singular" : "post.likeCount.plural"
```
bnolens, line 138: Wondering if this logic applies to all languages

bnolens:
Note to also translate strings in our info.plist. Thinking of the remote-notification warning for instance. Also App Store descriptions and In-app purchase labels (in IAPConfig).
What about push notification? Our NotificationService assumes notifications are in English.

bnolens:
Users should be able to change the app language in their settings. Does this Preferred Language setting work? 

https://github.com/TheBLVD/mammoth/assets/221925/304e8b51-e875-4cd7-90a7-6187882b1cf9

derspyy:
> What about push notification? Our NotificationService assumes notifications are in English.
> 
> 

hmm tough. we could either rework that, or move notifications to use, like, emojis? instead of text?

Mammoth/Views/Cells/PostCardCell/PostCardMetadata.swift
derspyy: worst case, they'll have an equivalent string for both, no? i mean, unless we change the number representation, in which case we just make that "1" a localized string.

derspyy:
> Users should be able to change the app language in their settings. Does this Preferred Language setting work?
> 
>  RPReplay_Final1709033609.MP4

is this already present? we'll need to go over these settings.

bnolens:
> > Users should be able to change the app language in their settings. Does this Preferred Language setting work?
> > RPReplay_Final1709033609.MP4
> 
> is this already present? we'll need to go over these settings.

It is already present yes

bnolens:
I rebased from main, enabled localization in Interface Builder files (for the onboarding) and added a JumpToNow translation key. 

https://github.com/TheBLVD/mammoth/assets/221925/3bc221de-7857-4277-85cb-8617611c69bc

bnolens (changes requested):
I still see a considerable amount of strings not yet converted to NSLocalizedString. Not saying we should be close to 100% now but we're not yet in a place that can be shipped tbh. It's easy to test by forcing a non-translated language in the simulator and see which strings are not displayed as keys. See video. 

https://github.com/TheBLVD/mammoth/assets/221925/2dfea1ed-49ea-4332-b543-08b2dd8b6513

derspyy:
accidental force-push erasure. closing.

PR #435, opened 2024-02-28
MAM-3869 Fix bookmark and favourite list scrolling

nickludlam:
- Added `NewsFeedViewModel.makePaginatedRequest` to determine which feed types make paginated requests.
- Changed bookmarks and favourites to use `fetchAllPaginated()`
- Added `nextPageRange` and `previousPageRange` to NewsFeedViewModel to store our position in the paginated data
- Making paginated requests for unsupported feedTypes will throw the new `NewsFeedViewModelError.invalidFeedType`
- Removed cache hydration for all paginated feed types, as we cannot locate you correctly within the pages without storing the previous/next responses from the server. This intervention is made in `NewsFeedViewModel.syncDataSource()`. We still persist and retrieve the data, it is just ignored for now.
- Added a missing call to `shouldPollForListData` before `startPollingListData()`
- Added bookmarks to the exclude list of `shouldPollForListData` and `shouldSyncItems`
- One or two other minor fixes

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewModel.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewModel.swift
@@ -8,6 +8,10 @@
 
 import UIKit
 
+enum NewsFeedViewModelError: Error {
+    case invalidFeedType
+}
+
 enum NewsFeedSnapshotUpdateType {
     case hydrate        // loading items from cache
     case replaceAll     // replacing all items
@@ -133,12 +137,10 @@ enum NewsFeedTypes: CaseIterable, Equatable, Codable, Hashable {
                 return (result.enumerated().map({ .postCard(PostCardModel(status: $1, withStaticMetrics: false, instanceName: $1.serverName, batchId: batchName, batchItemIndex: $0)) }), cursorId: cursorId)
                 
             case .likes:
-                let (result, cursorId) = try await TimelineService.likes(range: range)
-                return (result.enumerated().map({ .postCard(PostCardModel(status: $1, withStaticMetrics: false, batchId: batchName, batchItemIndex: $0)) }), cursorId: cursorId)
-                
+                throw NewsFeedViewModelError.invalidFeedType
+
             case .bookmarks:
-                let (result, cursorId) = try await TimelineService.bookmarks(range: range)
-                return (result.enumerated().map({ .postCard(PostCardModel(status: $1, withStaticMetrics: false, batchId: batchName, batchItemIndex: $0)) }), cursorId: cursorId)
+                throw NewsFeedViewModelError.invalidFeedType
                 
             case .mentionsIn:
                 let (result, cursorId) = try await TimelineService.mentions(range: range)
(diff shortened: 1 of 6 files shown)


nickludlam:
This should reference MAM-3869

nickludlam:
Things I'd like to improve:

- Retain the ability to use cached data, but always start you scrolling from the top
- Sync properly with any additions or removals you might make on the server outside of Mammoth (i.e. using the web client)
- Once you scroll to the bottom of the list, we disable the "pull up to load more" functionality, as the server has specifically told us no more information will be available. This is not the same as timeline data.

Mammoth/Screens/HomeScreen/NewsFeedViewModel.swift
```diff
@@ -382,6 +405,15 @@ class NewsFeedViewModel {
         }
     }
     
+    internal var makePaginatedRequest: Bool {
```
bnolens, line 408: If all fetch requests return a `Paginated` object (as proposing in my other comment) the NewsFeedViewModel should not be aware of which request is paginated and which is not, it just sends the Paginated object and the fetch request knows how to handle it.

Mammoth/Screens/HomeScreen/NewsFeedViewModel.swift
```diff
@@ -181,6 +183,24 @@ enum NewsFeedTypes: CaseIterable, Equatable, Codable, Hashable {
         }
     }
     
+    func fetchAllPaginated(range: RequestRange = .default, batchName: String, retryCount: Int = 0) async throws -> ([NewsFeedListItem], Pagination?, String?) {
```
bnolens, line 186: If all fetch requests return a `Paginated` object, you don't need to split the fetch function in 2 anymore

bnolens (changes requested):
Awesome work! I left a few comments. I think there's a way you could simplify the code a little bit by having all requests return the same pagination object or enum, instead of the `cursorId` we currently are using. That way the NewsfeedViewModel stays a bit dumber and the code could be simplified a bit. 

This might increase a little bit the scale of this project, but it is a key part of the app so I think we need to be sure to tackle it the right way.

Mammoth/Services/Backend/TimelineService.swift
```diff
@@ -97,16 +97,17 @@ struct TimelineService {
         return result
     }
     
-    static func likes(range: RequestRange = .default) async throws -> ([Status], cursorId: String?) {
+    static func likes(range: RequestRange = .default) async throws -> ([Status], Pagination?, cursorId: String?) {
         let request = Favourites.all(range: range)
-        let result = try await ClientService.runRequest(request: request)
-        return (result, cursorId: result.last?.id)
+        let (result, pagination) = try await ClientService.runPaginatedRequest(request: request)
+        return (result, pagination, cursorId: result.last?.id)
```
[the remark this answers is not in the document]
nickludlam: Yeah, initially I had a method which responds with no cursorId, but I was having issues with the way in which cursorId is used to control scrolling and fetches, and it looked sufficiently complicated to make me want to avoid too much disruption, so I adopted it's continued use.

I can look separately at where and how cursorId is used by the NewsFeedViewModel/Controller and at least propose some further changes if not make the changes as part of this PR.

nickludlam:
I'm not as familiar with the Mastodon API to necessarily know many of the endpoints can take pagination parameters, or respond with `max_id`/`min_id` headers.

The other thing we ought to plan for in this feedback is the future way in which we could use and sync offline data.

PR #437, opened 2024-02-29
Fix issue with the '@' prefix not being added appropriately in Mention function

nickludlam:
On a user profile page, if you select "Mention" from the top right menu, it incorrectly misses out the '@' prefix for the full username. I've added it, plus changed the comments in UserActions to be clearer about what is happening.

### DIFF
--- a/Mammoth/Utils/UserActions.swift
+++ b/Mammoth/Utils/UserActions.swift
@@ -132,7 +132,7 @@ struct UserActions {
         let vc = NewPostViewController()
         vc.isModalInPresentation = true
         vc.fromPro = true
-        vc.proText = "\(user.remoteFullOriginalAcct) "
+        vc.proText = "@\(user.remoteFullOriginalAcct) "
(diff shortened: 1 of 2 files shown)


Mammoth/Utils/UserActions.swift
```diff
@@ -132,7 +132,7 @@ struct UserActions {
         let vc = NewPostViewController()
         vc.isModalInPresentation = true
         vc.fromPro = true
-        vc.proText = "\(user.remoteFullOriginalAcct) "
+        vc.proText = "@\(user.remoteFullOriginalAcct) LETS GOOOO"
```
bnolens, line 135: Probably didn't want to commit LETS GOOOO?

bnolens, line 135: Probably didn't want to commit LETS GOOOO?

nickludlam: Haha oops yes! I'll fix that

nickludlam:
Updated with the right changes!

PR #471, opened 2024-03-07
MAM-4012+4013+4017

derspyy:
had to do a single PR because or else i'd have to wait for merge bc of conflicts and all that.
also single commit because xcode crashed rendering localizable.xcstrings.


bnolens (approved):
🙌

PR #472, opened 2024-03-08
MAM-4018-client-side-fallback-to-only-enable-supported-languages

bnolens:
This is an important step in our l10n workflow. With these changes we need to explicitly enable a language in our codebase. Even if translations on crowdin are approved and downloaded in our xcstring files, a language will only be used if it is in the `supportedLocalizations` array (https://github.com/TheBLVD/mammoth/compare/MAM-4018-client-side-fallback-to-only-enable-supported-languages?expand=1#diff-1e390366736fb502afd833b0a67be240fc83822ed4b39b3e8bf332839df7def8R26)

This will function as an extra step in enabling new languages, which gives us more control. 
This will allow us to easily create (Testflight) builds with specific languages enabled or disabled, or have a system enabling languages for specific users.

### DIFF
--- a/.arkana.yml
+++ b/.arkana.yml
@@ -9,6 +9,7 @@ global_secrets: # Optional. List of secrets that are the same regardless of whic
   - SwiftyGiphyAPI
   - IAPVerificationSecret
   - JoinCommunityPageURL
+  - CrowdinDistributionString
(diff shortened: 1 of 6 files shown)


derspyy:
oh, awesome. : ) lgtm

derspyy (approved):
LGTM

PR #479, opened 2024-03-09
MAM-3857-redundant-excessive-custom-emoji-image-requests

derspyy:


Mammoth/Models/PostCardModel.swift
```diff
@@ -807,7 +807,10 @@ extension PostCardModel {
     
     func preloadEmojis() {
         self.emojis?.forEach({
-            ImageDownloader.default.downloadImage(with: $0.url)
+            if !SDImageCache.shared.diskImageDataExists(withKey: $0.url.absoluteString) {
+                let prefetcher = SDWebImagePrefetcher.shared
+                self.imagePrefetchToken = prefetcher.prefetchURLs([$0.url], context: [.animatedImageClass: SDAnimatedImageView.self], progress: nil)
```
bnolens, line 812: Reusing the same `imagePrefetchToken` prop probably not what we wanna do here. 
It's getting reset in each for each iteration and also the prefetrchURL call is an async call and the token should say available until the request completed. Because we want to cancel the request early in some scenarios. 
Wondering if you should store the token in the first place for these emojis?

Mammoth/Models/UserCardModel.swift
```diff
@@ -244,7 +244,10 @@ extension UserCardModel {
         }
         
         self.emojis?.forEach({
-            ImageDownloader.default.downloadImage(with: $0.url)
+            if !SDImageCache.shared.diskImageDataExists(withKey: $0.url.absoluteString) {
```
bnolens, line 247: Same comment about the token here

bnolens (changes requested):
Nice improvement!!! Please have a look at my comment.

Mammoth/Models/PostCardModel.swift
derspyy: would it work if i just kept that same download function as before? is that part of the same cache?

bnolens (approved):
LGTM

PR #481, opened 2024-03-10
MAM-3955-blurhash-images-have-wrong-width-in-carousel

derspyy:
this should be good enough to balance performance and precision.


bnolens (approved):
LGTM

Issue #365, opened 2024-01-26
Audio Support


ringmaster:
### Is your feature request related to a problem? Please describe.

I am not able to post audio from my phone, nor am I able to play back audio that has been posted via the Mastodon web client. 

### Describe the solution you'd like.

I would like to be able to select audio files from Mammoth to post to my Mastodon instance. I would also like to be able to click the playback button on audio files in Mammoth and hear audio along with operating playback controls. 

### Describe alternatives you've considered.

I posted audio to Mastodon via the web UI, but then when I returned to Mammoth, the file would not play. It just showed a non-working play button. 

### Additional context.

See https://mastodon.social/@ringmaster/111824271565558044 for a port with audio.
![This is what it looks like in Mammoth, and it does not play](https://github.com/TheBLVD/mammoth/assets/127738/4a6e0ea8-6c12-42ca-8ca3-eb96f3bcf4f7)

derspyy:
will close this issue because the audio files are playing correctly, but i'll start an internal discussion about adding a built-in audio recorder, upload, player, etc.

PR #484, opened 2024-03-10
MAM-4032-listening-to-audio-not-working

derspyy:
part of #365, also.

### DIFF
--- a/Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift
+++ b/Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift
@@ -454,8 +454,9 @@ final class PostCardVideo: UIView {
                 
                 loadingIndicator.startAnimating()
                 
-                let previewImageURL = URL(string: media.previewURL ?? "")
-                previewImage.sd_setImage(with: previewImageURL)
+                if let previewImageURL = media.previewURL{
+                     previewImage.sd_setImage(with: URL(string: previewImageURL))
+                }
                 
                 if let cachedPlayer = cachedPlayer {
                     // if the player is preloaded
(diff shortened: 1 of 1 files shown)


Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift
```diff
@@ -450,9 +450,11 @@ final class PostCardVideo: UIView {
             guard self.media != media else { return }
             
             self.media = media
-            if let videoURL = URL(string: media.remoteURL ?? media.url), let previewImageURL = URL(string: media.previewURL ?? "") {
+            if let videoURL = URL(string: media.remoteURL ?? media.url) {
                 
                 loadingIndicator.startAnimating()
+                
+                let previewImageURL = URL(string: media.previewURL ?? "")
```
bnolens, line 457: I'd say, unwrapping the optional like in this example would be cleaner
```
if let previewImageURL = URL(string: media.previewURL) {
     previewImage.sd_setImage(with: previewImageURL)
}

**Review by bnolens — CHANGES_REQUESTED** [review]
Please have a look at my comment

File: Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift (hunk 1 above)
Comment (derspyy) [reply Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift:457]: you're right.

Comment (derspyy) [reply Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift:457]: well, actually, 

```
if let previewImageURL = media.previewURL{
     previewImage.sd_setImage(with: URL(previewImageURL))
}
```

bnolens: Yes that works too

bnolens (approved):
LGTM!

PR #485, opened 2024-03-10
MAM-3990-images-in-akkoma-posts-have-wrong-dimensions

derspyy:
(there is an issue with the diff where an indentation change looks horrible)

this addresses the issue where some services omit image metadata (due to not doing background processing).
affects: akkoma users & mastodon users viewing akkoma profiles or replies.

we need to see if this breaks anything, i made images without metadata fallback to 16:9 with a fit scale. this is how they look:

image

this is how they compare to akkoma's frontend:

image

mastodon images should be untouched.


Mammoth/Views/Cells/PostCardCell/PostCardImage.swift
```diff
@@ -285,57 +285,57 @@ final class PostCardImage: UIView {
...
-            if self.media?.meta?.original?.aspect == nil {
-                self.media?.meta?.original?.aspect = Double(self.media?.meta?.original?.width ?? 10) / Double(self.media?.meta?.original?.height ?? 10)
+            // meta itself might be nil
+            var aspect: Double? = nil
+            if let width = self.media?.meta?.original?.width, let height = self.media?.meta?.original?.height {
+                aspect = Double(width) / Double(height)
             }
+            let ratio = self.media?.meta?.original?.aspect ?? aspect ?? 16.0 / 9.0
     
-            if let ratio = self.media?.meta?.original?.aspect {
```
derspyy, line 293: just noticed that the diff here is horrible. i just changed this first line, the rest is indentation.

Mammoth/Views/Cells/PostCardCell/PostCardVideo.swift
```diff
@@ -518,58 +518,58 @@ final class PostCardVideo: UIView {
...
-                if self.media?.meta?.original?.aspect == nil {
-                    self.media?.meta?.original?.aspect = Double(self.media?.meta?.original?.width ?? 10) / Double(self.media?.meta?.original?.height ?? 10)
+                // meta itself might be nil
+                var aspect: Double? = nil
+                if let width = self.media?.meta?.original?.width, let height = self.media?.meta?.original?.height {
+                    aspect = Double(width) / Double(height)
                 }
+                let ratio = self.media?.meta?.original?.aspect ?? aspect ?? (self.media?.type == .audio ? 1.0 : 16.0 / 9.0)
         
-                if let ratio = self.media?.meta?.original?.aspect {
```
derspyy, line 526: same here

bnolens (changes requested):
There's an issue where the blurhash image is not the same size as the final image. If that's fixed, I think looks good. 👍 
https://github.com/TheBLVD/mammoth/assets/221925/ee33b753-957c-4b78-80fb-ffc6c8aca467

derspyy:
i really don't think this is related to my pr (which only concerns images with no metadata and therefore no blurhash), but i can also fix that if you want

bnolens:
> i really don't think this is related to my pr (which only concerns images with no metadata and therefore no blurhash), but i can also fix that if you want

What about this change you made? https://github.com/TheBLVD/mammoth/pull/485/files#diff-db60f1901e92e5866bdbe16565649da7b8f6e6b3c4888169a3ac93faad313f6cR26

derspyy:
> > i really don't think this is related to my pr (which only concerns images with no metadata and therefore no blurhash), but i can also fix that if you want
> 
> 
> 
> What about this change you made? https://github.com/TheBLVD/mammoth/pull/485/files#diff-db60f1901e92e5866bdbe16565649da7b8f6e6b3c4888169a3ac93faad313f6cR26

for reference: since i changed the images to FIT instead of FILL, the placeholders with wrong dimensions became more visible. fixed it though

derspyy:
made it so that only these images will be set to FIT scale.

bnolens (approved):
LGTM

PR #486, opened 2024-03-11
MAM-4034-cant-interact-with-remote-posts-in-akkoma

derspyy:


Mammoth/Services/Backend/StatusService.swift
```diff
@@ -495,10 +495,9 @@ extension StatusService {
         } catch let error {
             do {
                 switch error as? ClientError {
-                case .mastodonError(let message):
-                    if message == "Record not found" {
-                        return try await self.runTaskLocally(forStatus: status, task: task)
-                    }
+                case .mastodonError:
+                    // these error messages can be localized.
```
bnolens, line 499: Are there other mastodon errors than the `Record not found` one? I would guess there are. 
In this scenario, we only want to run the task locally if the post is not found. You must keep this check - by looking at an error code instead? Or is there an error key in this error object we could use?

bnolens (commented):
Please have a look at my comment

Mammoth/Services/Backend/StatusService.swift
derspyy: we could look at the 404 itself. wdyt?

bnolens: Might work yes. Is a `Record not found` always paired with a 404? Can 404 happen in other scenarios?
@rileyhCode do you think checking the 404 error code is enough here?

rileyhCode: Hm. It looks like "Record not found" is coming from the data returned (MastodonError type, json "error" field).
Anyway… checking for 404 would probably work (but should be tested, obv…)

derspyy: not an easy way to do this using the current library...

derspyy:
not ready for merge BTW

PR #489, opened 2024-03-12
MAM-4036-selecting-german-isnt-showing-german-localization

derspyy:


rileyhCode:
I wonder if we need a more general solution. For example, I found these on the web (related to iOS):
de-AT for German as used in Austria
de-CH for German as used in Switzerland

Should those use the current 'de' localization? I don't know enough about these regions to answer.

bnolens:
> I wonder if we need a more general solution. For example, I found these on the web (related to iOS): de-AT for German as used in Austria de-CH for German as used in Switzerland
> 
> Should those use the current 'de' localization? I don't know enough about these regions to answer.

Good point @rileyhCode . We should include all german dialects because, I believe, they're close enough to each other. @derspyy Could you come up with a change to include all DE dialects?

bnolens (approved):
LGTM

PR #490, opened 2024-03-12
MAM-3626-consider-use-url-search-on-instance-to-get-quote-post-from

derspyy:
allows us to render every fediverse object as a quote post, rather than only mastodon stuff, also as a bonus makes quoting conform to the instance's moderation, since the user's instance fetches the post.

this is mainly a POC, currently incompatible with ivory because they just append the link under the post.

images: (can i make these smaller?)
image
image

### DIFF
--- a/Mammoth/Managers/StatusCache.swift
+++ b/Mammoth/Managers/StatusCache.swift
@@ -125,59 +125,23 @@ class StatusCache {
             return
         } else {
             // Make the network request, then the callback
-            guard let server = url.host else {
-                completion(url, nil)
-                return
-            }
-            
-            // Special case for threads.net posts
-            if server.hasSuffix("www.threads.net") {
-                Task {
-                    let baseURL = "https://feature.moth.social"
-                    let request = URLRequest(url: URL(string: "\(baseURL)/api/v1/status/lookup?url=\(url.absoluteString)")!)
-                    do {
-                        let (data, response) = try await URLSession.shared.data(for: request)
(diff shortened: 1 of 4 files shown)


Mammoth/Managers/StatusCache.swift
```diff
@@ -125,59 +125,23 @@ class StatusCache {
...
+            Task {
+                do {
+                    let result = try await ClientService.runRequest(request: request)
+                    if let stat = result.statuses.first {
+                        await MainActor.run {
                             StatusCache.cachedStatusesLock.lock()
                             StatusCache.cachedStatuses[url] = stat
                             StatusCache.cachedStatusesLock.unlock()
                         }
+                        completion(url, stat)
```
bnolens, line 138: Please add a `log.error("couldn't find post.")` statement or something similar

derspyy: will do

derspyy:
image

i've seen this case where the url card seems to take preference, but it can't say if this is particularly _wrong_ behaviour. do we want to overwrite these?

derspyy:
image

this is also very useful to view quotes from akkoma -> mastodon. this is how they'd look originally:

image

bnolens (approved):
Great work @derspyy. LGTM! I made one small change to limit the QuotePost search to only 1 result.

derspyy:
> Great work @derspyy. LGTM! I made one small change to limit the QuotePost search to only 1 result. 

cool, no problem

PR #497, opened 2024-03-18
MAM-4022-feed-stability-work

bnolens:
Storing deferred snapshot updates in an array and executing them when scrolling ended was probably a bad design. When calling the deferred snapshot updates the scroll position was often not retained, which led to feed jumps.

In this PR, we don't store deferred snapshot updates but instead apply the latest snapshot changes once, when scrolling stops.

https://linear.app/theblvd/issue/MAM-4022/feed-stability-work


derspyy:
great solution! 🔥

derspyy:
not entirely familiar, but nothing jumped out immediately. i say we test this to see if jumps persist

PR #507, opened 2024-03-19
MAM-4048-double-tap-home-is-unreliable

bnolens:
Double-tapping the home tab now has the same behavior as the JumpToNow pill. When tapped, the feed clears and the newest content is displayed. 

https://linear.app/theblvd/issue/MAM-4048/double-tap-home-is-unreliable


bnolens:
Self-approving this one but feel free to leave comments retroactively.

PR #509, opened 2024-03-20
MAM-4048-double-tap-home-is-unreliable

bnolens:
[MAM-4048 : Double tap home is unreliable](https://linear.app/theblvd/issue/MAM-4048/double-tap-home-is-unreliable)


bnolens:
Self-merging this one, but feel free to comment

PR #515, opened 2024-03-20
MAM-3695-repeating-poll-options

derspyy:


Mammoth/Views/Cells/PostCardCell/PostCardPoll.swift
```diff
@@ -100,7 +100,7 @@ extension PostCardPoll {
     func configure(postCard: PostCardModel) {
         self.postCard = postCard
         
-        if let poll = postCard.poll {
+        if let poll = postCard.poll, optionsStackView.arrangedSubviews.isEmpty {
```
bnolens, line 103: The one regression with this approach is if the poll votes change between the feed load and the post sync the view will not update. It will still update if you're the one voting.

derspyy: should i then clear the subviews instead?

bnolens: Clearing subviews and rebuilding the UI in the configure method turned out to be bad for performance. The right way to fix this is to build the poll once in `setupUI` and only adjust the UI in `configure`. It's better not to adding or removing child views in `configure`.  Updating the view's `.isHidden` property and NSLayoutContraint constants is a better approach.

bnolens: Clearing subviews and rebuilding the UI in the configure method turned out to be bad for performance. The right way to fix this is to build the poll once (in `setupUI` if possible) and only adjust the UI in `configure`. It's better not to add or remove child views in `configure`. Updating the view's `.isHidden` property and NSLayoutContraint constants is a better approach.

derspyy: got it. awesome

derspyy: what i still need to figure out is how to initialize it with the post data, haha 😅

Mammoth/Views/Cells/PostCardCell/PostCardPoll.swift
```diff
@@ -77,11 +77,30 @@ class PostCardPoll: UIView {
 // MARK: - Setup UI
 private extension PostCardPoll {
     func setupUI() {
+        self.isHidden = true
         self.isOpaque = true
         self.addSubview(mainStackView)
         
         mainStackView.addArrangedSubview(optionsStackView)
-        
+        if let postCard = self.postCard {
```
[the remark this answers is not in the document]
derspyy: oh, yep, this is some test code that i left by mistake.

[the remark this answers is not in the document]
bnolens: This poll component is still very expensive to render because we remove subviews in `prepareForReuse`. I'm going to merge this PR since it appears to solve the repeating poll option bug but we'll have to look into making this view more performant. We cannot remove views and recreate views each time this view is reused.

bnolens (approved):

PR #517, opened 2024-03-21
MAM-4062-voting-on-polls-from-other-instances-fix

bnolens:
https://linear.app/theblvd/issue/MAM-4062/voting-on-polls-from-other-instances-fix


derspyy:
i was considering a solution with FetchPolicy, does this work when a poll is fetched from a remote profile?

bnolens:
> i was considering a solution with FetchPolicy, does this work when a poll is fetched from a remote profile?

This is supposed to work in all scenarios. It does also do a search of the post if it comes from the same instance, so in that case it's overkill, but since it only happens with a user votes I think it's ok.

derspyy:
cool

PR #529, opened 2024-03-24
MAM-4071-interactions-going-up-but-not-back-down

derspyy:
sorry for the previous oversight 😅

### DIFF
--- a/Mammoth/Models/PostCardModel.swift
+++ b/Mammoth/Models/PostCardModel.swift
@@ -888,22 +888,22 @@ extension PostCardModel {
 extension PostCardModel {
     
     public func likeTap() -> Void {
-        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited) == false else { return }
+        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited ?? false) == false else { return }
         (status.reblog ?? status).likeTap()
     }
     
     public func unlikeTap() -> Void {
-        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited) == true else { return }
+        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited ?? false) == true else { return }
         (status.reblog ?? status).unlikeTap()
     }
     
     public func repostTap() -> Void {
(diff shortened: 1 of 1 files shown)


Mammoth/Models/PostCardModel.swift
```diff
@@ -888,22 +888,22 @@ extension PostCardModel {
...
     
     public func unlikeTap() -> Void {
-        guard case .mastodon(let status) = data, status.favourited == true else { return }
+        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited) == true else { return }
         (status.reblog ?? status).unlikeTap()
     }
     
     public func repostTap() -> Void {
-        guard case .mastodon(let status) = data, status.reblogged == false else { return }
+        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited) == false else { return }
```
bnolens, line 901: This line is supposed to be `guard case .mastodon(let status) = data, (status.reblog?.reblogged ?? status.reblogged ?? false) == false else { return }`

Mammoth/Models/PostCardModel.swift
```diff
@@ -888,22 +888,22 @@ extension PostCardModel {
...
         (status.reblog ?? status).repostTap()
     }
     
     public func unrepostTap() -> Void {
-        guard case .mastodon(let status) = data, status.reblogged == true else { return }
+        guard case .mastodon(let status) = data, (status.reblog?.favourited ?? status.favourited) == true else { return }
```
bnolens, line 906: This line is supposed to be `guard case .mastodon(let status) = data, (status.reblog?.reblogged ?? status.reblogged ?? false) == true else { return }`

bnolens (changes requested):
Great catch @derspyy. Please review my comment for an additional fix.

bnolens:
I pushed the additional changes in https://github.com/TheBLVD/mammoth/pull/529/commits/4ee1f36ed8804a905dbaa6e863e83d74edcb77be

bnolens (approved):
LGTM

Mammoth/Models/PostCardModel.swift
derspyy: can you plz make these changes? can't rn...

bnolens: Already did

PR #536, opened 2024-03-28
Update MastodonKit.

derspyy:
some changes to comply with mastodon updates. 
you can use these to see some changes we'd have to do to the ui.

for example:
- `indexable` and `hide_collections` option in profile editor.
- `admin` and `severed_relationships` notifications support.
- lists `exclusive` option.


bnolens (approved):
LGTM

PR #538, opened 2024-03-31
MAM-4087-recursive-quote-posts-have-link-card

derspyy:
my simple solution was adding a little indicator that a quote post has another quote post. thought it would be simpler than making them render recursively or something.

i tried making it look like the "quote post not found" screen.

image

> says "quotes another post" in portuguese.


bnolens (approved):
LGTM

PR #540, opened 2024-04-03
MAM-4093-replies-arent-being-threaded

derspyy:


Mammoth/Screens/Composer/NewPostViewController.swift
```diff
@@ -3759,6 +3759,43 @@ class NewPostViewController: UIViewController, UITableViewDataSource, UITableVie
...
+                        log.error("error from Search.search(): \(error)")
+                        // I have seen 500, 503 errors returned when the serer is very busy
+                    }
+                    if let results = statuses.value {
+                        let statuses = results.statuses
+                        if let statID = statuses.first?.id {
+                            successGettingPostID = true
+                            DispatchQueue.main.async {
+                                // Try again
+                                self.postNextThreadPiece(postPieces, inReplyTo: statID)
```
bnolens, line 3781: Using `self` in a closure can cause reference cycles so you should use `[weak self] in` at the top of the closure on line L3767

Mammoth/Screens/Composer/NewPostViewController.swift
```diff
@@ -3759,6 +3759,43 @@ class NewPostViewController: UIViewController, UITableViewDataSource, UITableVie
...
+                            }
+                        } else {
+                            log.error("Expected a status")
+                        }
+                    }
+                    // Put an alert to retry if needed.
+                    if !successGettingPostID {
+                        DispatchQueue.main.async { [weak self] in
+                            self?.setPostFailure()
```
bnolens, line 3790: Idem

bnolens (changes requested):
Added a small comment to review

derspyy:
> Added a small comment to review

are you able to make these changes? won't be available until the late afternoon, so it may not be ready for release.

bnolens:
> > Added a small comment to review
> 
> are you able to make these changes? won't be available until the late afternoon, so it may not be ready for release.

Sure. Will do

bnolens (approved):

PR #541, opened 2024-04-03
MAM-4084-urls-on-the-composer-should-always-be-23-characters-in

derspyy:
the `postPiecesFromPost()` rewrite should be a WAY saner solution, that also accounts for URLs.
what it does is iterate word for word and push a new piece only if the current word would exceed the limit.


bnolens (approved):
LGTM

PR #550, opened 2024-04-10
MAM-4105-fix-for-post-card-header-height-issue-in-full-mode

bnolens:
https://linear.app/theblvd/issue/MAM-4105/fix-for-post-card-header-height-issue-in-full-mode


bnolens:
Simple fix - self-merging

PR #551, opened 2024-04-10
MAM-4107-fix-client-issue-with-localized-instances-list

bnolens:
https://linear.app/theblvd/issue/MAM-4107/fix-client-issue-with-localized-instances-list

### DIFF
--- a/Mammoth/Screens/Registration/SignInViewController.swift
+++ b/Mammoth/Screens/Registration/SignInViewController.swift
@@ -128,12 +128,11 @@ class SignInViewController: UIViewController, UITableViewDataSource, UITableView
     }
     
     class func loadInstances(isFromSignIn: Bool) {
-        let urlStr = "https://feature.\(GlobalHostServer())/api/v1/instances/list"
+        let urlStr = "https://feature.moth.social/api/v1/instances/list"
         let url: URL = URL(string: urlStr)!
         var request = URLRequest(url: url)
         var components = URLComponents(url: request.url!, resolvingAgainstBaseURL: false)!
-        let language = NSLocale.preferredLanguages[0]
-        if GlobalStruct.supportedLocalizations.contains(language) {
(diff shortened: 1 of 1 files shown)


bnolens:
Auto-approving this one so I can spin a build. But let me know if you have comments.

Mammoth/Screens/Registration/SignInViewController.swift
```diff
@@ -128,12 +128,11 @@ class SignInViewController: UIViewController, UITableViewDataSource, UITableView
...
     
     class func loadInstances(isFromSignIn: Bool) {
-        let urlStr = "https://feature.\(GlobalHostServer())/api/v1/instances/list"
+        let urlStr = "https://feature.moth.social/api/v1/instances/list"
         let url: URL = URL(string: urlStr)!
         var request = URLRequest(url: url)
         var components = URLComponents(url: request.url!, resolvingAgainstBaseURL: false)!
-        let language = NSLocale.preferredLanguages[0]
-        if GlobalStruct.supportedLocalizations.contains(language) {
+        if let language = NSLocale.current.languageCode, GlobalStruct.supportedLocalizations.contains(language) {
```
derspyy, line 135: this fails because languageCode = "pt" and supportedLocalizations has "pt-BR" or whatever it looks like. it was correct.

derspyy (commented):
the part mentioned there should be reverted.

Mammoth/Screens/Registration/SignInViewController.swift
bnolens: Right. But the previous code didn't work either for all other languages.
For instance: I'm on "nl-BE" but supported languages only include "nl". I'll fix the "pt-BR" case.

bnolens: Fix in https://github.com/TheBLVD/mammoth/pull/552

PR #552, opened 2024-04-10
MAM-4108-make-instances-list-sorting-work-of-pt-br

bnolens:
https://linear.app/theblvd/issue/MAM-4108/make-instances-list-sorting-work-of-pt-br


derspyy (commented):
i don't understand why this was necessary (or if it's supposed to fix other stuff as well) if the previous check was working properly, and the issue was sorting.

derspyy:
ah, just looked at the other issue. LGTM!

PR #555, opened 2024-04-14
MAM-3914-dont-replace-every-post-data-when-fetching-for-interaction

derspyy:
i basically inverted our logic. instead of overwriting the `.data` parameter and keeping a backup in `.preSyncData`, i'm keeping `.data` the original data and only writing new data to `.remoteData`, which is only used for interactions and app name.

i think this is a good quick fix for weird ID behavior until we figure out a better solution.

### DIFF
--- a/Mammoth/Models/PostCardModel.swift
+++ b/Mammoth/Models/PostCardModel.swift
@@ -25,7 +25,7 @@ final class PostCardModel {
     }
     
     var data: Data
-    var preSyncData: Data?
+    var remoteData: Data?
     
     /// staticMetrics should be true when posts are coming from a pre-defined
     /// server, e.g. the For You feed. For these posts metrics will not update
@@ -65,7 +65,6 @@ final class PostCardModel {
     let rebloggerUsername: String
     var richRebloggerUsername: NSAttributedString?
     
-    var applicationName: String?
     let visibility: String?
     
     var account: Account?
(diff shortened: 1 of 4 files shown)


bnolens:
I like this direction. We should map data to its origin. So if a post comes from forYou or a SM (via moth.social) we should store the status data like that in the PostCardModel. If we sync the post with it's local data, we should store the local data separately in the PostCardModel. Maybe `data` in PostCardModel could be a dictionary. Let's fully map out what we want to do here and come up with a design that can scale to all of the above.

derspyy:
currently, data is only the originally fetched Status. so if the post was fetched from the FY or SL, `.data` is not from the user instance. are you suggesting that in those cases we should default to `.remoteData` and leave `.data` only for stuff from the user's instance?

derspyy:
what we could do is make `data` an optional `localData`, make two functions fetching one from the other and read both like `localData ?? remoteData!` (or the opposite depending in priority) whenever we need post data.

derspyy:
and then interactions are super simplified because now instead of retryLocally we can just read `localData` and fetch it if it's `nil`.

derspyy:
i'm thinking of this pr as a simple fix to those weird "id mismatch" issues before we think of an alternative

Mammoth/Models/PostCardModel.swift
```diff
@@ -700,57 +707,7 @@ final class PostCardModel {
...
-        }
-        // application name is only known by the original post
-        self.applicationName = ((newStatus.reblog?.application ?? newStatus.application)?.name.stripHTML() ?? newStatus.reblog?.application?.name ?? newStatus.application?.name)
-        
-        if self.applicationName == nil && newStatus.serverName == "www.threads.net" {
-            self.applicationName = "Threads"
-        }
-        
-        // Filters
-        self.filterType = newStatus.filtered?.reduce(FilterType.none) { result, current in
```
bnolens, line 735: Looks like you didn't reimplement this filterType logic.

Mammoth/Models/PostCardModel.swift
```diff
@@ -700,57 +707,7 @@ final class PostCardModel {
...
-            if case .hide(_) = result { return result }
-            if current.filter.filterAction == "hide" { return .hide(current.filter.title) }
-            return .warn(current.filter.title)
-        } ?? FilterType.none
-        
-        // Muted and Blocked
-        let blockedIds = ModerationManager.shared.blockedUsers.map { $0.remoteFullOriginalAcct }
-        let mutedIds = ModerationManager.shared.mutedUsers.map { $0.remoteFullOriginalAcct }
-        
-        if let acctID = self.account?.remoteFullOriginalAcct {
```
bnolens, line 745: Looks like the following logic is not re-implemented

Mammoth/Models/PostCardModel.swift
derspyy: we don't need it, since the remoteData doesn't have that field read. 
we only did that to preserve the original data, which i'm doing in a saner alternative.

Mammoth/Models/PostCardModel.swift
derspyy: idem.

bnolens (approved):
LGTM

Issue #562, opened 2024-04-27
Toggle to hide content that has been filtered does not work


SuperFola:
### Is your feature request related to a problem? Please describe.

I’m filtering a few hashtags and words, because I don’t want to hear from the bird site in my feeds. However on Mammoth there is a post with an overlay saying that something has been filtered.

### Describe the solution you'd like.

I’d like to be able to toggle this behavior on and off: show that something has been filter or just don’t show the filtered content.

derspyy:
hmm i think this may be residual from when filtering was done client-side... can you send a screenshot of that overlay?

SuperFola:
It looks like this when my Twitter filter hits:

![image](https://github.com/TheBLVD/mammoth/assets/10694085/92a6d292-a449-4d8e-93e4-de84f5656c56)

derspyy:
oh! this is a filter option from mastodon. you should be able to enable "Hide Completely" on the filter's options. : )

SuperFola:
Ah yes! The thing is, I already have "hide completely" on all my filters:

![image](https://github.com/TheBLVD/mammoth/assets/10694085/94e7b757-186c-4ea3-b678-a388d301874a)

SuperFola:
I didn't think of checking there too, but it's also on "hide completely" (with all the extras turned on)

derspyy:
i think the issue is that we're just not hiding the post when it has that flag. will totally be looking into that.

PR #565, opened 2024-05-06
MAM-4125-hide-completely-is-not-hiding-the-post

derspyy:
fixes an oversight:
we have 2 filter passes. one for unauthenticated requests (like the fy) and another for authenticated requests, and it seems like the the latter was not being applied correctly.

addresses #562 <3


Mammoth/Localizable.xcstrings
rileyhCode, line 1: Did you mean to include this change in the Localizable.xcstrings? (GitHub isn't showing me the specific diffs here…)

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
```diff
@@ -49,7 +49,7 @@ extension NewsFeedViewModel {
                     } else if case .channel = currentType {
                         newItems = items.removeMutesAndBlocks().removeFiltered()
                     } else {
-                        newItems = items
+                        newItems = items.removeFiltered()
```
rileyhCode, line 52: Based on the comment on line 43, it seems like mutes/blocks/filters are already expected to be applied to the feeds coming directly from the instance. If that's the case, we shouldn't need to apply the .removeFiltered() here.

Or is there something that is causing the filter to not apply on some feeds? Like… maybe the user changes the filter, but they already have consumed some posts, so we need to filter them locally?

(I suspect it's the latter case. If so, can you update the comment to explain why the .removeFiltered() is needed for feeds coming straight from the instance?)

rileyhCode (changes requested):
Pls review notes. Thx.

Mammoth/Localizable.xcstrings
derspyy: yes. i localized the filter overlay because that was missing.

Mammoth/Screens/HomeScreen/NewsFeedViewModel+Services.swift
derspyy: mastodon gives us the post with the "hide" tag instead of just ignoring it when it hits a hiding filter, which is the original oversight (possibly).

basically, that second filter pass just goes "if we have a 'hide' tag, ignore".

rileyhCode (approved):
OK; LGTM.

PR #566, opened 2024-05-08
MAM-4124-interaction-counts-removing-post-metadata

derspyy:
i see some weird visuals when unliking a post and those probably need to be fixed.


bnolens (approved):
LGTM

PR #576, opened 2024-05-28
MAM-4138-recursive-quote-indicator-above-images

derspyy:


Mammoth/Views/Cells/PostCardCell/PostCardQuotePost.swift
```diff
@@ -744,7 +744,7 @@ fileprivate class PostCardQuoteIndicator: UIStackView {
         self.spacing = 8.0
         
         self.isLayoutMarginsRelativeArrangement = true
-        self.directionalLayoutMargins = NSDirectionalEdgeInsets(top: 4, leading: 0, bottom: 0, trailing: 0)
+        self.directionalLayoutMargins = NSDirectionalEdgeInsets(top: 4, leading: 12, bottom: 0, trailing: 0)
```
bnolens, line 747: Is this intentional?

bnolens (changes requested):
Please remove my note

Mammoth/Views/Cells/PostCardCell/PostCardQuotePost.swift
derspyy: yes, it's because we're not using the contentStack padding.

bnolens: gotcha

bnolens (approved):
LGTM

PR #585, opened 2024-06-09
Project Configuration

dimitribouniol:
I noticed that:
a) we can define the ruby version locally to minimize version drift between developers, and 
b) ArkanaKeys was not part of the Xcode project, so I added it as a dependency.


Mammoth.xcodeproj/project.pbxproj
```diff
@@ -1372,6 +1373,7 @@
 		53B1585C293F66D20026E90B /* Decode85.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Decode85.swift; sourceTree = "<group>"; };
 		53B15868294227650026E90B /* FilterDetailsViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FilterDetailsViewController.swift; sourceTree = "<group>"; };
 		53EB77ED29D2EAC80036CDE1 /* NavigationHelpers.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NavigationHelpers.swift; sourceTree = "<group>"; };
+		655609B72C1757AA004233BC /* README.md */ = {isa = PBXFileReference; lastKnownFileType = net.daringfireball.markdown; path = README.md; sourceTree = "<group>"; };
```
dimitribouniol, line 1376: This makes accessing the README easy in Xcode. It's not added to the compiled binary:
image

README.md
```diff
@@ -11,20 +11,82 @@ The Mammoth Team
...
+```
+
+3. Initialize `rbenv` by installing it in your zsh profile, and reload the profile (this assumes you are using `zsh` by default — if you aren't, please run `rbenv init` instead and follow instructions):
+
+```
+% echo "eval \"\$(rbenv init - zsh)\"" >> ~/.zshrc
+% exec zsh
+```
+
+4. Install a compatible version of ruby, and continue the steps above:
```
rileyhCode, line 76: Should this be "…continue the steps below:"?

README.md
```diff
@@ -43,31 +105,11 @@ The Mammoth Team
 You must use Bundler 2 or greater with this lockfile.
 ```
 
-Ruby 3 is required. Check your ruby version `ruby --version`.
-
-1. install brew [(see web page instructions)](https://brew.sh)
-2. install rbenv
-   [(brew install rbenv ruby-build)](https://github.com/rbenv/rbenv/blob/master/README.md)
-3. run `rbenv init` and follow the printed instructions
-4. install current Ruby version `rbenv install 3.2.1`
```
rileyhCode, line 52: Even though it can be dicey to specify a version of ruby here, there is an implication that this is a "known working version". I wonder if we should either keep this instruction, or… at least add a comment indicating this is a known working version.

(though… above you set a version of 3.3.2 - so maybe that works?) Hm. Just trying to make it easiest for someone new coming onto the project.

rileyhCode (changes requested):
It's great to get some fresh eyes on the setup instructions. Pls review my comments, and then make changes and/or reply as you see fit. Thx!

README.md
dimitribouniol: No, above since you would need to bundle install and bin/arkana to finish the process

README.md
dimitribouniol: Yeah, I verified 3.3.2 is a known working version, so it shouldn’t be an issue from this point forward since rbenv will use that version since it’s part of the repo now — everyone will just end up using that version.

README.md
dimitribouniol: This now links back to the requirements section to make it extra clear!

rileyhCode (approved):
Looks good to me.

Issue #150, opened 2023-10-17
Range of photo sizes on timeline


justaname2:
**Is your feature request related to a problem? Please describe.**

As a photographer, I'm frustrated that photos on the timeline are so small. (Maybe they are OK on phones, but not on my iPad Pro 12.9".) They are so small as to have virtually no visual impact, so there's no reason for readers to click through

**Describe the solution you'd like**
Default photosize for posters to maximum width of the timeline, regardless of device, whether portrait or landscape (like Twitter), and don't crop (unlike Twitter). If some users don't like that, let them pick settings for full size, medium, and small in their Preferences.

**Additional context**
Photos and artwork should be first class citzens.

PR #589, opened 2024-06-17
Full Width Media and Posts

dimitribouniol:
I added an option to allow media such as photos, videos, and web previews to take up a much larger presence on screen for users who would prefer more visual content in their timelines rather than textual content. This should address #150.

https://github.com/TheBLVD/mammoth/assets/225505/3e6e4833-cb4d-4701-8e44-6a25ad8ecca2

https://github.com/TheBLVD/mammoth/assets/225505/4b38c6c0-739b-45ca-bb94-44fac67e5bda

### Todos and Followups in future PRs:
- [x] Make sure this works well on iPad and macOS — currently I only tested on iPhone.
- Make the profile picture smaller when the user info is a single line.
- Add maximum height constraints to prevent exceedingly tall images from scrolling forever.
- Make the first image of a carousel large to showcase it.
- Add a dedicated timeline to collapse textual content, showcasing media much more prominently.


Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
```diff
@@ -597,15 +619,33 @@ private extension PostCardCell {
...
+                    self.mediaContainerWrapper.bottomAnchor.constraint(equalTo: self.mediaContainer.bottomAnchor),
+                ])
+            } else {
+                NSLayoutConstraint.activate([
+                    self.mediaContainer.leadingAnchor.constraint(equalTo: self.mediaContainerWrapper.leadingAnchor),
+                    self.mediaContainerWrapper.trailingAnchor.constraint(equalTo: self.mediaContainer.trailingAnchor),
+                    self.mediaContainer.topAnchor.constraint(equalTo: self.mediaContainerWrapper.topAnchor),
+                    self.mediaContainerWrapper.bottomAnchor.constraint(equalTo: self.mediaContainer.bottomAnchor),
+                ])
+            }
```
rileyhCode, line 638: A couple of notes on this section with the constraints:
- It's a little confusing some are in the order of mediaContainer, mediaContainerWrapper and others are the reverse order (mediaContainerWrapper, mediaContainer). Consider picking one to be consistent.
- Consider adding a comment about the only difference being the leading constraint; otherwise, the next engineer will have to sift through to figure it out.

dimitribouniol: They are in this order: `b.anchor.constraint(equalsTo: a.anchor, constant: c)`, such that `b = a + c`, where `c` is positive, even if omitted. Having them be all on one side has always led to flipping it down the line in my experience, as you'd end up needing to either make `c` negative or flip the anchors. This also has the side benefit that it helps to indicate which view is a child of the other when used consistently.

If you'd rather I change it so things align instead, I can, but al least now you know _why_ I'd opt for this version instead 😅 

(This practice comes from [Apple's long-lost guidance](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/AutolayoutPG/AnatomyofaConstraint.html): "Positive constants are favored over negative constants")

I'll be sure to add some comments indicating the differences though!

Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
```diff
@@ -679,6 +720,13 @@ private extension PostCardCell {
             self.webview?.isHidden = true
             mediaContainer.addArrangedSubview(self.webview!)
             webviewTrailingConstraint = self.webview!.trailingAnchor.constraint(equalTo: mediaContainer.layoutMarginsGuide.trailingAnchor)
+            
+            /// Adjust the immersive images to hand over their container
+            if self.cellVariant.mediaVariant == .immersive, let first = self.mediaContainer.arrangedSubviews.first {
+                let leadingConstraint = self.mediaContainer.leadingAnchor.constraint(equalTo: first.leadingAnchor, constant: 60)
+                leadingConstraint.priority = .required
+                NSLayoutConstraint.activate([leadingConstraint])
+            }
```
dimitribouniol, line 729: I don't think this worked, I need to remove it

Mammoth/Views/Cells/PostCardCell/PostCardCell.swift
```diff
@@ -688,10 +735,10 @@ private extension PostCardCell {
         contentWarningButton.isHidden = true
         contentWarningButton.addTarget(self, action: #selector(self.contentWarningButtonTapped), for: .touchUpInside)
         contentWarningConstraints = [
-            contentWarningButton.topAnchor.constraint(equalTo: header.bottomAnchor, constant: -1),
+            textAndSmallMediaStackView.topAnchor.constraint(equalTo: contentWarningButton.topAnchor, constant: 0),
             contentWarningButton.bottomAnchor.constraint(equalTo: contentStackView.bottomAnchor, constant: -48),
-            contentWarningButton.leadingAnchor.constraint(equalTo: contentStackView.leadingAnchor, constant: 6),
-            contentWarningButton.trailingAnchor.constraint(equalTo: contentStackView.trailingAnchor, constant: 3),
+            textAndSmallMediaStackView.leadingAnchor.constraint(equalTo: contentWarningButton.leadingAnchor, constant: 3),
+            contentWarningButton.trailingAnchor.constraint(equalTo: textAndSmallMediaStackView.trailingAnchor, constant: 3),
```
dimitribouniol, line 741: This is now a bit cleaner for all scenarios:
https://github.com/TheBLVD/mammoth/assets/225505/95458514-60bb-42a6-b7a9-e85da0ab7ebf

derspyy (approved):

PR #591, opened 2024-06-19
feat(subclub): v1 integration.

derspyy:
- [x] detect SC accounts and main accounts linking SC.
- [x] show subscribe button on both accounts.
- [x] show subscription post tab on main acc.
- [x] detect a SC redirect.
- [x] polish subscription flow.
- [x] fix issue with ProfileHeader not being updated.

after rebasing, a new issue was introduced where clicking a username in a user's profile fields fetches the wrong acc.

- [ ] fix that


Mammoth/Models/UserCardModel.swift
```diff
@@ -166,6 +171,16 @@ class UserCardModel {
         self.fields = account.fields
         self.fields?.forEach({$0.configureMetaContent(with: emojisDic)})
         self.joinedOn = account.createdAt?.toDate()
+        self.tippable = account.acct.hasSuffix("social-proxy.com")
```
bnolens, line 174: Would be good to put this string in a constant so it's easy to update when we change domains. Maybe in .env?

Mammoth/Models/PostCardModel.swift
```diff
@@ -316,6 +316,15 @@ final class PostCardModel {
         return sourceDescription
     }
     
+    // check if the post is from a tip account.
+    var tipAccount: Bool {
```
bnolens, line 320: Our booleans in here mostly start with `is`. Would make sense to use `isTipAccount` here too

Mammoth/Models/UserCardModel.swift
```diff
@@ -67,6 +67,9 @@ class UserCardModel {
         return ModerationManager.shared.blockedUsers.first(where: {$0.remoteFullOriginalAcct == self.uniqueId}) != nil
     }
     
+    let tippable: Bool
```
bnolens, line 70: Idem (`isTippable`)?

Mammoth/Models/UserCardModel.swift
```diff
@@ -166,6 +171,16 @@ class UserCardModel {
         self.fields = account.fields
         self.fields?.forEach({$0.configureMetaContent(with: emojisDic)})
         self.joinedOn = account.createdAt?.toDate()
+        self.tippable = account.acct.hasSuffix("social-proxy.com")
+        // detect tippable link in profile fields.
+        var acct: String? = nil
+        for field in account.fields {
+            if let s = field.value.matchingStrings(regex: "https://social-proxy.com/users/([a-z0-9-_]+)").first?[1] {
```
bnolens, line 178: Item, don't hardcode the social-proxy url

Mammoth/Screens/ProfileScreen/Views/ProfileHeader.swift
```diff
@@ -374,31 +401,41 @@ extension ProfileHeader {
...
+        // add subscribe button if:
+        // 1. user is a tippable account and isn't already subcribed to.
+        // 2. user has a tippable account linked.
+        // not add button on own account.
+        if ((user.tippable && user.followStatus != .following) || user.tippableAccount != nil) && !user.isSelf {
+            tipButton.isHidden = false
+            tipButton.addTarget(self, action: #selector(self.subscribeTapped), for: .touchUpInside)
+        }
+        
         let joined_on = user.joinedOn?.toString(dateStyle: .short, timeStyle: .none) ?? ""
```
bnolens, line 439: Not your change, but this variable should be camelcase. You mind changing it?

Mammoth/Screens/ProfileScreen/Views/ProfileHeader.swift
```diff
@@ -578,6 +615,22 @@ extension ProfileHeader {
         }
     }
     
+    @objc func subscribeTapped() {
+        triggerHapticImpact(style: .light)
+        if let user = user {
+            switch user.tippable {
+            case true:
+                if let url = URL(string: "https://social-proxy.com/subscribe/to/\(user.username)?amount=500&currency=USD") {
```
bnolens, line 623: Item, try not to hardcode the domain

bnolens: The new url is actually `https://social-proxy.com/@\(user.username)/subscribe?amount=500&currency=USD`, please update to that. The old one will still work for a while though

Mammoth/Screens/ProfileScreen/Views/ProfileHeader.swift
```diff
@@ -578,6 +615,22 @@ extension ProfileHeader {
...
+                    PostActions.openLink(url)
+                }
+            case false:
+                if let username = user.tippableAccount, let url = URL(string: "https://social-proxy.com/subscribe/to/\(username)?amount=500&currency=USD") {
```
bnolens, line 627: Item url changes

bnolens (changes requested):
Great! Please check my notes

bnolens:
Wondering how well the following status gets updates after returning from the social-proxy.com review. Is it always correctly saying "Following"? If not, we should force a relationship sync when coming back.

bnolens:
I just realized this is still a draft. I might have been a bit quick in reviewing this, but progress looks good!

derspyy:
> Wondering how well the following status gets updates after returning from the social-proxy.com review. Is it always correctly saying "Following"? If not, we should force a relationship sync when coming back.

agreed. this is part of the subscription flow that i'm still planning on polishing.

derspyy:
also agreed on not hardcoding the domain. i'll work on that

derspyy:
update: i still need to test the following status update. i couldn't because of the issues when subscribing.

Mammoth/Screens/ProfileScreen/Views/ProfileHeader.swift
```diff
@@ -618,13 +619,13 @@ extension ProfileHeader {
...
-            switch user.tippable {
+            switch user.isTippable {
             case true:
-                if let url = URL(string: "https://social-proxy.com/subscribe/to/\(user.username)?amount=500&currency=USD") {
+                if let url = URL(string: "https://" + ArkanaKeys.Global().subClubDomain + "/@\(user.username)/subscribe?callback=mammoth://social-proxy=\(user.username)&amount=500&currency=USD") {
                     PostActions.openLink(url)
                 }
             case false:
-                if let username = user.tippableAccount, let url = URL(string: "https://social-proxy.com/subscribe/to/\(username)?amount=500&currency=USD") {
+                if let username = user.tippableAccount, let url = URL(string: "https://" + ArkanaKeys.Global().subClubDomain + "/@\(username)/subscribe?callback=mammoth://social-proxy=\(username)&amount=500&currency=USD") {
```
bnolens, line 628: This path should be `https://subclub.social/@premiumUsername/subscribe?callback=mammoth://social-proxy&id=accountUsername&amount=500&currency=USD`

reference in icecubes: https://github.com/Dimillian/IceCubesApp/commit/3fbfecb171a4c12a78bafbd7f280b188239a2353#diff-3e9e838a1912ac03de6d850047cf6f7188e589934090657edafefc581bc5c850R126

derspyy: i added the username to the callback so that we could use it to follow the correct account right away

derspyy: (take a look at the SceneDelegate impl)

bnolens: Oh I see, great. 
But there's still an issue with the usernames you pass. The username after the @ (premium usename) should be the acct of the creator, so without `@subclub.social`. 
You should also pass the full acct handle of the currently logged-in account as a URL parameter with key `id`. (see example above)

derspyy:
tomorrow i'll check tippableAccount to see if i can store the domain separately from the username, because i need it for the callback...

derspyy:
(adding the domain and then removing it right afterwards is just a quick hack i made this morning for it to work, i'll probably end up storing the username by itself and just adding it back in the callback

PR #592, opened 2024-06-20
Add SwiftLint to project

billburgess:
I added SwiftLint support to this project. I opted for Homebrew installation as we are already using that for other dependencies and is the easiest for developers that are getting setup on the project.

This PR is best reviewed commit by commit as there are a lot of files with changes. I grouped the commits logically so all of the same changes were made in the same commit for easier readability.

Finally, we can haggle over the settings and how strict we want things to be going forward. I borrowed this SwiftLint config file from other projects I've used which was provided to the open source community to use. It's well documented and easy to understand which rules are in use and which are not.

Initial install of SwiftLint before any config changes:
Screenshot 2024-06-20 at 8 29 01 AM

After addition of config file:
Screenshot 2024-06-20 at 8 32 34 AM

Finally, after ignoring items we don't care about:
Screenshot 2024-06-20 at 8 47 46 AM

All breaking issues have been resolved. There are still many warnings throughout the app that can be addressed at a later time.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -652,10 +652,10 @@ opt_in_rules: # some rules are only opt-in
   - legacy_multiple  
 
 excluded: # paths to ignore during linting. Takes precedence over `included`.
+  - bin
+  - ArkanaKeys
   - MammothTests
   - MammothUITests
-  - ArkanaKeys
-  - bin
 
 custom_rules:
   force_https:
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,672 @@
+disabled_rules: # rule identifiers to exclude from running
+
+  # Rationale: Xcode auto indentation can cause this warning
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#closure-end-indentation
+  - closure_end_indentation
+
+  # Rationale: Arbitrary restriction
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#cyclomatic-complexity
+  - cyclomatic_complexity
+
+  # Rationale: Arbitrary restriction
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#discouraged-object-literal
+  - discouraged_object_literal
+
+  # Rationale: Current implementation triggers on system APIs. Disabling until rule is configurable to fix this. An optional collection can have two "empty" states: nil and empty. If nil and empty are to mean different things, consider modeling it differently using an enum with associated values.
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#discouraged-optional-collection
+  - discouraged_optional_collection
+
+  # Rationale: Doesn't really add any clarity
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#explicit-acl
+  - explicit_acl
+
+  # Rationale: Heavy handed in telling someone how to code
+  # https://github.com/realm/SwiftLint/blob/master/Rules.md#explicit-enum-raw-value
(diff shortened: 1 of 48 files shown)


README.md
```diff
@@ -98,7 +104,7 @@ ruby 3.2.1 (2023-02-08 revision 31819e82c8) [arm64-darwin22]
   the repo.
   ![XcodeArkanaBuildError](https://github.com/TheBLVD/mammoth-app/assets/76360/ec0fd8a9-285f-41dd-817d-60fc41d94e54)
 
-- If you run `bin/arckan` and see the following:
+- If you run `bin/arkana` and see the following:
```
bnolens, line 107: 👍

Mammoth.xcodeproj/project.pbxproj
```diff
@@ -4496,6 +4519,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 468;
 				DEVELOPMENT_TEAM = W44P88K5SB;
+				ENABLE_USER_SCRIPT_SANDBOXING = NO;
```
bnolens, line 4522: Is this a swiftlint requirement?

bnolens (commented):
Any rules you think should be enabled but didn't to limit the diff?

billburgess:
@bnolens No, I disabled the 2 inline and left everything else. They are still there and will show up as compiler warnings. These came from some friends in the open source community and I've had good luck with them. They usually only need minor changes to get things right.

Mammoth.xcodeproj/project.pbxproj
billburgess: Yes. Starting in Xcode 15 this defaults to YES and forces scripts into a sandbox. It needs to be set to no to allow SwiftLint to run. If we are worried about it or would like to keep sandboxing on, I can try to find a workaround. I'm guessing SwiftLint is touching files outside the sandbox which triggers the warning.

billburgess: tl;dr; We can't install SwiftLint via SPM and the project is a little less safe as a result.

I tried adding this via SPM to start with, but ran into issues. MessageKit has a version of the SwiftLintPlugin as a part of that and our import fails when we try to bring it in solo. I couldn't figure a way around that. None of the other methods of getting the plugin worked for me other than Pods (which we didn't have set up) or home-brew. I do think that this makes the project a little less "safe" by having sandboxing turned off. But we do code reviews before merging and should be able to control anything that could run outside the sandbox.

Ideally we'd have the plugin come from SPM to reduce dependencies on the developer from wanting to build and first run the project. There are plenty of other setup tasks already that kind of makes this a moot point. I'll defer to the groups wisdom on what to do here.

.swiftlint.yml
```diff
@@ -0,0 +1,672 @@
+disabled_rules: # rule identifiers to exclude from running
```
bnolens, line 1: Aren't these rules all disabled? I'm wondering if we should enable some of these (in the future)

bnolens (approved):
LGTM

.swiftlint.yml
billburgess: Yes, the .swiftlint.yml has all of the rules listed by default. This file is explicit by disabling rules and enabling others. If you aren't explicit, it will use the default value. I'm very much in favor of adding any of these we may want in the future.

Mammoth.xcodeproj/project.pbxproj
bnolens: Seems ok for now, I agree

PR #594, opened 2024-06-21
Add haptics for home menu items

billburgess:
I don't have a way to test this at the moment since you need developer account access to add physical devices. I think this is the right change based on what I found in the other context menu items.

I added haptic impact light to the home menu items as well as a missing one from the profile context menu. There were also a few lines of spaces cleaned up. Otherwise, minor changes here. Please let me know if this is every spot for this. I'm trying to have a small feedback when the long press presents this menu:

Screenshot 2024-06-20 at 7 35 17 PM

It already presents a feedback only when tapping the Add account button. I added one for changing accounts here:

Screenshot 2024-06-20 at 7 36 21 PM

If I'm missing one, which I think I might be, let me know and I can update to add it in the correct spot.


bnolens (approved):
LGTM!

Issue #582, opened 2024-06-09
Hide compose button


Zabriskije:
### Is your feature request related to a problem? Please describe.

I find the compose button distracting.

### Describe the solution you'd like.

I'd like to have the option to hide the compose button. That could be added to the list visible with a long press on the user icon:

![hide compose](https://github.com/TheBLVD/mammoth/assets/101254295/34fa895d-bf03-4b39-b448-1ef86f3c2256)

PR #599, opened 2024-06-27
Screens: Show and hide New Post button on scroll

awilfox:
The button hides when the user is scrolling to view older posts, and reappears when the user is scrolling to view newer posts.

Ref: #582


awilfox:
I have pushed a commit that also animates the change.

PR #601, opened 2024-06-30
Localization Sort Order

dimitribouniol:
Ever since I touched the localization file in #589, Xcode has complained that it is not in order and insists on re-writing it. This  sets the order, which should prevent issues going forward!


bnolens:
Note that this file is overwritten by a generated version by Crowdin. We use a Github action to create PRs with the updated version pulled from Crowdin. https://github.com/TheBLVD/mammoth/actions/workflows/crowdin-download.yml

dimitribouniol:
We should make sure that action also writes in sorted order — I'll look into it!

dimitribouniol:
Might just need to update to a newer version of the action, unless there are other configuration files at play: https://community.crowdin.com/t/apple-xcstrings-string-catalog-format-discrepancy-between-crowdin-and-xcode/5937

derspyy (approved):

PR #603, opened 2024-06-30
Extension Version Mismatch

dimitribouniol:
Moved versions to be defined on the project rather than individual targets so they stay in sync:

image
image
image

This prevents issues like this from cropping up by having a single place to make changes:

image


derspyy (approved):
super helpful : )

PR #613, opened 2024-07-15
MAM-4161-hide-reader-view-option-when-using-external-browser

bnolens:
- renaming browser option "Mammoth" to "In-app browser"
- only showing "reader view" option when "In-app browser" is being used
- making "in-app browser" the default


bnolens:
I'm self-approving this one, but feel free to review it and leave comments. I'll address them in a separate PR

inb4ohnoes:
Ideally we'd like to maintain consistency in capitalization, so I think it should be "In-App Browser" instead?

PR #616, opened 2024-07-15
MAM-4157-full-width-quote-posts-have-no-images-attached

derspyy:


inb4ohnoes (approved):
Tested on device, lgtm!

PR #617, opened 2024-07-16
MAM-4159: Improve UX around large media uploads

inb4ohnoes:
This improves upon the large media uploading experience. Mammoth now actively checks if large media has finished processing serverside.


derspyy (approved):
lgtm!

PR #618, opened 2024-07-16
PATCH-MAM-4161 browser selection

inb4ohnoes:
Fixes broken behavior from MAM-4161


derspyy:
don't we want this to be a localized string?

inb4ohnoes:
I believe ideally yes, but none of the other browser options are localized, including "System Default". I think that particular change is out of scope for this pr, so I've made another ticket for it at MAM-4166

PR #619, opened 2024-07-17
MAM-4156: Add special case for full width for details cells

inb4ohnoes:
Don't show full width layout in details so as not to break the threading UI


inb4ohnoes:
self-approving in the interest of timing

PR #626, opened 2024-08-07
MAM-4184-mentions-in-bio-get-removed-after-edit

derspyy:


inb4ohnoes (approved):
Tested adding a mention to test account bio then editing it again, seems to be fixed. Good catch!

PR #634, opened 2024-09-09
Add iCloud Sync

inb4ohnoes:
Improves on Bill's iCloud sync PR:

* Implements scrolling
* Saves the full model and offset to sync
* Saves keyed data based on account username
* Accounts for if the user happens to be scrolling
* Accounts for when data processing/saving is happening
* Accounts for "live syncing" timeline across multiple devices (i.e you have Mammoth open on two devices on the same feed, it will sync back and forth... with a small delay)
* Accounts for some other UX gotchas
* DOES NOT account for animating the scrolling, as that triggers some feed weirdness that's not really in scope to fix

Note: Sync seems to not work properly on first launch on simulator only. Physical device seems ok. Must have iCloud enabled to work. Sim -> Device seems to work, Device -> Device seem to work, Device -> Sim DOES NOT seem to work well.


rileyhCode (approved):
LGTM

PR #636, opened 2024-09-11
Use preUpdateScrollPosition

rileyhCode:
If I'm reading this correctly, sometimes the new scrollPosition was being used in the place of the pre-updateScrollPosition.

I gave the pre-updateScrollPosition its own name, and used it where it made sense based on the comments.


inb4ohnoes (approved):
👍

PR #637, opened 2024-09-11
Skip iCloud sync if user has manually scrolled

rileyhCode:
- add a userHasScrolledManually: Bool flag to newsFeedViewModel
- set userHasScrolledManually to true in NewsFeedViewController.scrollViewWillBeginDragging()
- have scrollToCloudPosition() bail if this is set to true

### DIFF
--- a/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
+++ b/Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
@@ -1287,6 +1287,8 @@ private extension NewsFeedViewController {
 // MARK: - Jump to newest
 extension NewsFeedViewController: JumpToNewest {
     func jumpToNewest() {
+        self.viewModel.userHasScrolledManually = true
+
         if !self.viewModel.pollingReachedTop {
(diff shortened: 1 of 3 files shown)


Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
```diff
@@ -728,6 +728,11 @@ extension NewsFeedViewController {
         viewModel.cancelPreloadCards(atIndexPaths: indexPaths)
     }
     
+    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
```
inb4ohnoes, line 731: My concern with this is does it get set in the same way if a user double taps the home button or if they tap status bar to scroll to top?

rileyhCode: Good point. If the user double taps the home button or taps the status bar to scroll to top, it should behave as if they had scrolled, and not cloud sync.

inb4ohnoes (approved):
🙏

PR #639, opened 2024-09-12
Update NewsFeedViewController.swift

rileyhCode:
- handle case of nil from cacheScrollPosition
- account for additional 16 pixels when using the custom CarouselNavigationHeader
- additional log


Mammoth/Screens/HomeScreen/NewsFeedViewController.swift
```diff
@@ -1247,7 +1250,16 @@ private extension NewsFeedViewController {
                     return nil
                 }
                 let rectForTopRow = tableView.rectForRow(at: currentCellIndexPath)
-                let offset = rectForTopRow.origin.y - pointWhereNavBarEnds.y
+                // The Mentions and Activity views use a custom CarouselNavigationHeader,
```
inb4ohnoes, line 1253: good catch

inb4ohnoes (approved):

